PR #1, opened 2024-05-03
Feature: add sentry.io

noahjoeris:
This PR adds sentry.io to the project. 
It includes error tracking, user bug report, and the sentry testkit.
In the future, the used account and organization needs to be updated.


app/sentry.client.config.ts
```diff
@@ -0,0 +1,35 @@
+// This file configures the initialization of Sentry on the client.
+// The config you add here will be used whenever a users loads a page in their browser.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from "@sentry/nextjs";
+
+Sentry.init({
+  dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
+
+  // Adjust this value in production, or use tracesSampler for greater control
```
NunoAlexandre, line 10: It would be good to know what the units of this rate is 👍

app/next.config.mjs
```diff
@@ -1,4 +1,44 @@
+import { withSentryConfig } from "@sentry/nextjs";
 /** @type {import('next').NextConfig} */
 const nextConfig = {};
 
-export default nextConfig;
+export default withSentryConfig(
```
NunoAlexandre, line 5: I guess this is a convention but I don't really want us to need a `withXYZ` every time we need to extend the config. Can't we just add the sentry stuff to the `nextConfig`? or have a `sentryConfig = {}` and then have a `fullConfig = { nextConfig, sentryConfig, ...}` gathering them all?

app/src/app/sentry-example-page/page.jsx
```diff
@@ -0,0 +1,79 @@
+"use client";
```
NunoAlexandre, line 1: Do we really need this page?

app/.env.local.example
```diff
@@ -0,0 +1,2 @@
+# Provide your Sentry DSN to allow Sentry to capture errors in your application.
+NEXT_PUBLIC_SENTRY_DSN=''
```
NunoAlexandre, line 2: GH shows there's no newline at the end of this file which will most likely be automatically added by most editors, so let's just add it already to avoid newline diffs/commits later on 👍

NunoAlexandre, line 2: Regarding the `NEXT_PUBLIC_SENTRY_DSN`, don't we need to set this already for debugging?

NunoAlexandre (commented):
Awesome :) Left some notes 👀

app/sentry.client.config.ts
noahjoeris: This file was auto generated by sentry. We can look into specific configs once we fully utilize it.

app/src/app/sentry-example-page/page.jsx
noahjoeris: No we don't. Was also auto generated and is a good way to test that it works. Especially once we switch the sentry organization. After that I would remove it anyways.

app/.env.local.example
noahjoeris: It is set in the `.env.local` file or in Vercel. This is an example environment file to know what env variables need to be defined. I do not want our Sentry DSN to be inside the repo.

NunoAlexandre: ah ok that makes sense 👌 The _public_ DSN should be fine to have on the repo, but having it on Vercel is good 👍

app/next.config.mjs
NunoAlexandre: Yeah that's fine 👍

app/sentry.client.config.ts
NunoAlexandre: I am trying to avoid having to look stuff up when having to change in the future, as most doc comments should aim to do.

It's explained here: https://docs.sentry.io/platforms/javascript/performance/#configure

> For example, to send 20% of transactions, set tracesSampleRate to 0.2.

NunoAlexandre, line 10:
```suggestion
  // The percentage of transactions that get sent, defined between 0 and 1. 
  // For example, to send 20% of transactions, set tracesSampleRate to 0.2
```

NunoAlexandre (approved):
Left a suggestion but no blockers so leaving it up to your judgment 👍

PR #5, opened 2024-05-13
Adds basic components to the app

noahjoeris:
This pull request adds a simple user interface. It allows to select a source and destination chain, a token, the token amount to transfer, and a receiver address.

**_Note: This is the first draft of the UI so it is not complete on purpose. And it does not include any functionality yet!_**

### Added Components
- `TransferPage:` The main component that glues everything together.
- `CustomSelect:` Button to select a chain or a token. Chain and token selection could be separated in the future if needed. But for now they are almost identical.
- `CustomSelectDialog:` The modal to show all the options to select from.
- `CustomSelectOption:` One option that can be selected.
- `AppHeader:` Simply the header above the transfer form.
- `ValueInput:` To input the token amount.
- `AddressInput:` To input the receiver address. 
- `WalletConnectButton:` Button to connect the source wallet.
- `TransferButton:` Button that will trigger the transfer.

### Others
- separates landing page from app
- fixes some husky pre-commit issues
- adds a basic model for `token` and `chain`
- uses framer motion as animation library

image


app/.gitignore
```diff
@@ -37,3 +37,6 @@ next-env.d.ts
 
 # Sentry Config File
 .sentryclirc
+
+# SWC Plugins directory
+.swc/
```
NunoAlexandre, line 42: Why are adding swc?

app/package.json
```diff
@@ -42,10 +33,9 @@
     "eslint-config-next": "14.2.3",
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-tsdoc": "^0.2.17",
-    "husky": "^9.0.11",
+    "husky": "^8.0.0",
```
NunoAlexandre, line 36: why the downgrade?

app/src/app/app/layout.tsx
```diff
@@ -0,0 +1,12 @@
+export const metadata = {
```
NunoAlexandre, line 1: What is this file? and why do we have an `app` dir within the `app` dir?  👀

NunoAlexandre (commented):
Some preliminary remarks & questions.

Regarding

> separates landing page from app

I don't like the new setup with `./app/app`. My idea initially was to have the landing page in the root dir and the app in the `app/` dir; the reason for that being that the landing page could be a statically generated site and cached heavily with a CDN and the app itself would be under a `app.` cname (app.turtle.cool) and would like to the vercel hosting.

I'm fine with combining the two things but then we should flatten the `app` dir in the root directory or discuss other alternatives. What are your thoughts on that?

app/src/components/AddressInput.tsx
```diff
@@ -0,0 +1,46 @@
+'use client'
+import { FC } from 'react'
+import { twMerge } from 'tailwind-merge'
+
+interface AddressInputProps {
+  /** The current input value, which should reflect the address. */
```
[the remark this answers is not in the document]
noahjoeris: Using `/** ...*/` is the correct TSDoc syntax. Either we use that or leave them out completely I would say.
The advantage is that, if you hover over the variables in vscode you see the doc comment in the context menu. Makes it easier and more reliable to use components.

app/src/app/app/global-error.tsx
```diff
@@ -10,7 +8,7 @@ interface GlobalErrorProps {
 
 const GlobalError: React.FC<GlobalErrorProps> = ({ error, reset }) => {
   useEffect(() => {
-    Sentry.captureException(error)
+    // Sentry.captureException(error)
```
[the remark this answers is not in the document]
noahjoeris: Yep woops. Got spammed with emails on every error so I commented it out 😆

app/src/app/app/layout.tsx
noahjoeris: I separated the landing page from the actual app. Therefore, we have two `layouts.tsx`, one for landing and one for the app. For me this was the easiest way to split them inside our nextjs app. But I am open for other approaches or other directory names.

And yeah as you said, I think the most common way to host is to put the transfer app on `app.turtle.cool` and the landing page on `turtle.cool`. Although, for this we probably need to put the app and landing into separate repos and have more hosting overhead. 

I am open for both ways 🚀

app/.gitignore
noahjoeris: Nextjs added a `.swc/...` directory on compiling the app and we don't want this in our repo right? I am not an swc expert.

app/src/components/AddressInput.tsx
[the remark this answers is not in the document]
NunoAlexandre: I agree the docs are nice to have. Bummer that the worst possible syntax was chosen for that tho 😆

app/src/app/app/global-error.tsx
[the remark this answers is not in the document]
NunoAlexandre: ahah ok. We can add a flag so that we don't get emails on development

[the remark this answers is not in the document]
NunoAlexandre: Created an issue for that, I might take it on me the coming days 👍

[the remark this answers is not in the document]
NunoAlexandre: @noahjoeris you can uncomment this on this PR but you can then always comment it for you locally and use git `update-index --assume-unchanged src/file/to/ignore` until we have the flag work done.

app/src/components/CustomSelect.tsx
```diff
@@ -0,0 +1,72 @@
...
+  /** Currently selected value, or null if no value is selected. */
+  value: SelectOption | null
+  /** Callback function that is invoked when the selected value changes. */
+  onChange: (newValue: SelectOption | null) => void
+  /** Array of options that the user can select from. */
+  options: SelectOption[]
+  /** Title of the select input, displayed when no item is selected. */
+  title: string
+  /** Whether the select input is disabled (non-interactive). */
+  disabled?: boolean
```
NunoAlexandre, line 21: I wonder if we would benefit from un-boolean this and make it a `Status`, like:

```
enum Status = Disabled | Enabled | etc
```

app/src/components/CustomSelect.tsx
```diff
@@ -0,0 +1,72 @@
+'use client'
+import { Chain } from '@/models/chain'
+import { Token } from '@/models/token'
+import Image from 'next/image'
+import { FC, useState } from 'react'
+import { twMerge } from 'tailwind-merge'
+import CustomSelectDialog from './CustomSelectDialog'
+
+export type SelectOption = Chain & Token
```
NunoAlexandre, line 9: How does this work? Doesn't this mean that `SelectOption` can therefore represent both a `Chain` and a `Token`, at the same time? We don't want that to be the case for the same component. Can't we make `CustomSelectProps` have a generic type param `SelectOption` instead?

app/src/components/TransferPage.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { Chain } from '@/models/chain'
+import { Token } from '@/models/token'
+import { FC, useState } from 'react'
+import AddressInput from './AddressInput'
+import CustomSelect from './CustomSelect'
+import TransferButton from './TransferButton'
+import ValueInput from './ValueInput'
+import WalletConnectButton from './WalletConnectButton'
+
+const TransferPage: FC = () => {
```
NunoAlexandre, line 12: Let's name this just as `Transfer` or something like that, since in the future we want to make this component embeddable by other projects in their own UI and "page" suggests that it's a separate screen

NunoAlexandre (commented):
A few comments but overall looks good :) happy to approve once I get your thoughts on the points

app/src/components/CustomSelect.tsx
noahjoeris: I get your point about boolean blindness, but the `disabled` boolean prop is such a common pattern that I am against this type of change. Every library contains this as a boolean so it would be more confusing than helpful. 
It is just for preventing state changes by disabling the `onChange` calls. 
But I am open for example use cases to extend it and to reconsider.

app/src/components/CustomSelect.tsx
noahjoeris: As we have not defined the Chain and Token interfaces yet, I did it this way. So SelectOption has all properties of Chain and Token. It's true it's only a short term approach.

And you are right a better solution is:
- We create a generic interface `SelectOption` and create helper functions to convert a `Chain` or a `Token` to `SelectOption`.

app/src/components/CustomSelect.tsx
NunoAlexandre: I get your point. We have to be sensible but also thoughtful. Just because something is a convention doesn’t mean it has to define the way we write code. This is not a big issue so I’m fine with going ahead like this 👍

noahjoeris:
I added a selectOption type. Let me know what you think about this approach

NunoAlexandre:
@noahjoeris thanks for the improvements!

Regarding `SelectOption` 👇 

Now we have the following:

1. `SelectOption` with `value`, `label`, and `logoURI`:
 
``` typescript
export interface SelectOption {
  /** The underlying value of the option. */
  value: Chain | Token
  /** The text to display for the option. */
  label: string
  /** The URI for the logo of the option. */
  logoURI: string
}
```

2. `Chain` and `Token` which basically copy the same fields, with the extra `symbol` on `Token`:

``` typescript
export interface Chain {
  id: string
  name: string
  logoURI: string
}

export interface Token {
  id: string
  name: string
  logoURI: string
  symbol?: string
}
```

With this setup, `SelectOption` by definition can only be used for `Chain` and `Token`s. What I was initially suggesting was that we would make `SelectOption` (or could be just `Option`) generic over the type of Option value, like:

``` typescript
export interface SelectOption<V> {
  /** The underlying value of the option. */
  value: V
  /** The text to display for the option. */
  label: string
  /** The URI for the logo of the option. */
  logoURI: string
}
```

This will of course require upstream changes to pass the specific type that's put up to choose from (now either `Chain` or `Token`).

This is not a blocker for me but we benefit from decoupling types with generic purposes (like an `Option`) from the specifics

app/src/models/selectOption.ts
```diff
@@ -0,0 +1,43 @@
...
+export interface SelectOption {
+  /** The underlying value of the option. */
+  value: Chain | Token
+  /** The text to display for the option. */
+  label: string
+  /** The URI for the logo of the option. */
+  logoURI: string
+}
+
+export const chainToSelectOption = (chain: Chain | null): SelectOption | null => {
```
NunoAlexandre, line 18: IMO we should not allow the param to be `null`, that should be filtered out already by the caller of `chainToSelectOption` (same for `tokenToSelectOption`). This function should always be a 1:1 and have no error cases, it's a pure function by definition.

NunoAlexandre (approved):
✨  Fantastic work! You rock, Noah 🙏

PR #8, opened 2024-05-21
Add evm and substrate wallet connection

noahjoeris:
This PR adds the functionality to connect substrate and evm wallets.

- I used Web3Modal by WalletConnect for EVM connections and Talisman Connect for Substrate connections. Talisman Connect is also a component that should make connecting wallets easy. But it needs to be said that it is very simplistic and has less features compared to Walletconnect. It is far from perfect but I researched alternatives and it is the best that works more or less out of the box and serves our purpose in the beginning. Maybe in the future Web3Modal will support Substrate chains. Or we switch to implementing it ourselves via useInkathon which is a great hook combining wallet and blockchain connection.
- WalletConnect does persist the wallet connection while Talisman Connect/PJS does not natively
    - Could be maybe implemented by storing it in local storage
- The new root folders `context` and `config` are a walletconnect convention according to the docs. I am open to restructure if needed.
- The buttons are intentionally not designed perfectly yet. I want to wait for the final design.
- The logic to switch wallet connections between source and destination is not included yet and I still need to think about how to do that.
- Best alternatives to Talisman Connect
  - [useInkathon](https://github.com/scio-labs/use-inkathon/tree/main) 
  - Plain Polkadot js
  - the new Polkadot API [still too early in dev, maybe in a couple of months] 
  - Web3Modal by WalletConnect [must use once Substrate support is released] 

image


noahjoeris:
@NunoAlexandre Can you check why the Vercel deployment fails? Locally it builds. Maybe due to our repo splitting?

app/src/components/ConnectSubstrateWalletButton.tsx
```diff
@@ -0,0 +1,71 @@
...
+}) => {
+  const [activeAccount, setActiveAccount] = useState<WalletAccount | null>(null)
+
+  // removes the active account if it is removed from the app
+  const handleUpdatedAccounts = (accounts?: WalletAccount[]) => {
+    if (!accounts || !activeAccount) return
+
+    for (let account of accounts) {
+      if (account.address === activeAccount.address) return
+    }
```
NunoAlexandre, line 31: this is fine but we could also `some` here:

```suggestion
    if (accounts.some(x => x.address === activeAccount.address)) { return }
```

app/src/components/Transfer.tsx
```diff
@@ -76,12 +78,13 @@ const Transfer: FC = () => {
...
-          <span className="label label-text">Receiver</span>
-          <AddressInput
+          <span className="label label-text">Receiver Address</span>
+          <ConnectSubstrateWalletButton label="Connect Substrate" />
+          {/* <AddressInput
             value={receiverAddress}
             onChange={setReceiverAddress}
             placeholder="Receiver Address"
-          />
+          /> */}
```
NunoAlexandre, line 87: to be removed?

app/src/config/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+// Get projectId at https://cloud.walletconnect.com
+export const projectId = process.env.NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID
+
+if (!projectId) throw new Error('Project ID is not defined')
+
+const metadata = {
+  name: 'Web3Modal',
+  description: 'Web3Modal Example',
+  url: 'http://localhost:3000', // origin must match your domain & subdomain
```
NunoAlexandre, line 14: will this work fine on Vercel?

NunoAlexandre, line 14: will this work fine on Vercel?

app/src/utils/address.ts
```diff
@@ -1 +1,7 @@
-// TODO: add address validation helper function
+// TODO: cleanup and make more readable.
+export const truncateAddress = (str = '', start = 4, end = 4) =>
```
NunoAlexandre, line 2: Please add a doc comment explaining what this is supposed to do 🙏

NunoAlexandre (commented):
Looking great overall, just a few comments 👍

app/src/utils/address.ts
noahjoeris: You're right, I forgot.

app/src/config/index.tsx
noahjoeris: Fair point. Updated it 👍

app/.husky/pre-commit
```diff
@@ -4,4 +4,5 @@
 cd "$(dirname -- "$0")/.."
 
 npm run format
+git add .
```
NunoAlexandre, line 7: Can you explain? I personally don't enjoy having git manipulation within a pre-commit hook but I'm open to understand your flow and maybe adjust to it 👍

app/src/utils/address.ts
```diff
@@ -1 +1,17 @@
-// TODO: add address validation helper function
+/**
+ * Truncate a blockchain address by showing the beginning and end parts.
+ *
+ * @param str - The address string to be truncated.
+ * @param start - Number of characters to show from the start. Default is 4.
+ * @param end - Number of characters to show from the end. Default is 4.
+ * @returns The truncated address or an empty string if the input string is empty.
+ */
```
NunoAlexandre, line 8: Thanks for adding this 🙏

app/.husky/pre-commit
noahjoeris: I noticed that formatting with prettier is otherwise not committed in some cases.
But it is more an edge case where a file is formatted that is not staged. Then this is not included in the commit.

But I do not mind removing it again.

NunoAlexandre: I see. Let's give it a try 👍 in this case, we could also add a `git commit -m "fmt"` command below to complete the flow

NunoAlexandre (approved):
❤️

noahjoeris:
Once we merge, our app on vercel will not work anymore. We need to add the WalletConnect ID to the env.

PR #10, opened 2024-05-27
Adds notifications to the app

noahjoeris:
This PR
- adds Zustand as a state management that globally stores notifications
- adds notification model, notification component, and a notification hook
- adds support for different notification severities (default, warning, info ...) that change the icon and background color. Also notifications support a header and a message.
- adds smooth fading in and out animations to the notifications

image

### DIFF
--- a/app/src/components/NotificationToast.tsx
+++ b/app/src/components/NotificationToast.tsx
@@ -1,7 +1,8 @@
 'use client'
-import { Notification } from '@/models/notification'
+import { Notification, NotificationSeverity } from '@/models/notification'
 import { motion } from 'framer-motion'
 import { useEffect } from 'react'
+import { twMerge } from 'tailwind-merge'
 
 const NOTIFICATION_TTL = 5000
 
@@ -14,6 +15,7 @@ const NotificationToast: React.FC<NotificationToastProps> = ({
   notification,
   removeNotification,
 }) => {
+  // Remove notification after TTL
   useEffect(() => {
     const timeoutRef = setTimeout(() => {
       removeNotification(notification.id)
@@ -22,6 +24,10 @@ const NotificationToast: React.FC<NotificationToastProps> = ({
     return () => clearTimeout(timeoutRef)
   }, [notification.id, removeNotification])
 
+  // Determine daisyui severity classname
+  const severityClass =
+    notification.severity !== NotificationSeverity.DEFAULT ? `alert-${notification.severity}` : null
+
   return (
     <motion.div
       layout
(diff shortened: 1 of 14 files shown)


app/src/components/NotificationToast.tsx
```diff
@@ -0,0 +1,150 @@
+'use client'
+import { Notification, NotificationSeverity } from '@/models/notification'
+import { motion } from 'framer-motion'
+import { useEffect } from 'react'
+import { twMerge } from 'tailwind-merge'
+
+const NOTIFICATION_TTL = 5000
```
NunoAlexandre, line 7: From a former colleague of mine :) https://ruudvanasseldonk.com/2022/03/20/please-put-units-in-names

app/src/components/NotificationToast.tsx
```diff
@@ -0,0 +1,150 @@
...
+    const timeoutRef = setTimeout(() => {
+      removeNotification(notification.id)
+    }, NOTIFICATION_TTL)
+
+    return () => clearTimeout(timeoutRef)
+  }, [notification.id, removeNotification])
+
+  // Determine daisyui severity classname
+  const severityClass =
+    notification.severity !== NotificationSeverity.DEFAULT ? `alert-${notification.severity}` : null
```
NunoAlexandre, line 29: Why do we need to handle the `DEFAULT` case differently? Can't we just make the call be the `notification.severity` and let the styling map what it should look like?

app/src/components/NotificationToast.tsx
```diff
@@ -0,0 +1,150 @@
...
+          fill="none"
+          viewBox="0 0 24 24"
+        >
+          <path
+            strokeLinecap="round"
+            strokeLinejoin="round"
+            strokeWidth="2"
+            d="M10 14l2-2m0 0l2-2m-2 2l-2-2m2 2l2 2m7-2a9 9 0 11-18 0 9 9 0 0118 0z"
+          />
+        </svg>
```
NunoAlexandre, line 111: Is it possible to host these svgs in a separate file? If we could do `return ./error.svg` that would be a lot cleaner. Just curious!

app/src/components/NotificationToast.tsx
NunoAlexandre: > I can try to pass in an unknown classname (alert-default). It could actually work and just be ignored. 👍

exactly, that's what I was thinking. and if we need a specific design for that we can always write our own design for that class, right?

app/src/components/NotificationToast.tsx
noahjoeris: Yep that's possible. Let's also not forget to let these alerts be styled by our designer.
So at best we have: info, warning, error, and success alerts.

app/src/components/NotificationToast.tsx
noahjoeris: correct

app/src/components/NotificationToast.tsx
noahjoeris: Not necessarily. It is just that daisyui has one.

NunoAlexandre: Ah ok, I oversaw that 👍

NunoAlexandre (approved):
🤟 awesome

PR #13, opened 2024-05-30
Add hooks useOngoingTransfers and useTransfer

noahjoeris:
This PR
- adds the hook `useOngoingTransfers` to handle transfers that are still in progress
- adds the hook `useTransfer` to handle the initiation of actual transfers. In the future it should figure out which services to call like AT or snowbridge etc.
- adds the service `ethereumTransfer` as an abstraction to use the snowbridge sdk
- adds the service `xcmTransfer` as an abstraction to use the AT API 

Note: `useTransfer`, `ethereumTransfer`, `xcmTransfer` act more like a template at the moment and still need to be implemented

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -38,6 +38,7 @@
   "devDependencies": {
     "@testing-library/jest-dom": "^6.4.5",
     "@testing-library/react": "^15.0.6",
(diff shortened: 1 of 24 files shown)


NunoAlexandre (approved):
LG 👍

PR #14, opened 2024-05-31
Integrate hooks and services

noahjoeris:
This PR
- integrates `useChains` and `useTransfer` hooks
- adds `chains` and `tokens` api endpoints 
- adds `useSubstrateWallet` hook to access the wallet signer


app/src/hooks/useTransfer.tsx
```diff
@@ -4,9 +4,17 @@ import { Token } from '@/models/token'
...
-  destChain: Chain
-  amount: bigint
-  recipient: string
+  destinationChain: Chain
+  amount: number
+  receiverAddress: string
+}
+
+interface TransferValidationParams {
+  token: Token | null
```
NunoAlexandre, line 13: out of curiosity, can't we use the `Type?` syntax?

app/src/hooks/useSubstrateWallet.tsx
```diff
@@ -0,0 +1,12 @@
+import { web3FromSource } from '@talismn/connect-components'
+
+const useSubstrateWallet = () => {
+  const getSigner = () => {
+    const injector = web3FromSource()
+    return injector.signer
+  }
+
+  return { getSigner }
```
NunoAlexandre, line 9: Why do we need this nested function definition?

NunoAlexandre, line 9: And shouldn't the `useSubstrateWallet` function return the signer type? I don't think I understand how this is meant to work 🤔

app/src/components/Transfer.tsx
```diff
@@ -84,9 +104,27 @@ const Transfer: FC = () => {
...
+            !isValid({
+              token,
+              sourceChain,
+              destinationChain,
+              amount,
+              receiverAddress,
+            })
+          }
+          onClick={() => {
+            if (sourceChain && destinationChain && token && amount)
```
NunoAlexandre, line 119: Isn't this `if` both redundant and erroneous? as in, the user can only call it if the button is not `disabled`, meaning it's only callable if it `isValid`. so checking if the values are set is not adding any extra benefit, right?

NunoAlexandre (commented):
some question 👀

app/src/hooks/useTransfer.tsx
NunoAlexandre: oh I see. Ok that makes sense 👍

NunoAlexandre: hm what about introducing an `Option` type? https://dev.to/martinpersson/a-guide-to-using-the-option-type-in-typescript-ki2

No blocker for me, but I haven't played with `nulls` in years so this caught my eye

app/src/hooks/useSubstrateWallet.tsx
noahjoeris: Welcome to the weird world of connecting substrate wallets. That's the DevEx you get by using talisman connect. Not comparable to Web3Modal.

Talisman connect puts the signer into local storage and this function reads from local storage. So the problem is signer could get outdated. 

Alternatively we could store it inside the app's state and trigger an update function every time the selected account changes. Might be better.

I still do not like Talisman connect and I don't know if we should use this long term.

app/src/components/Transfer.tsx
NunoAlexandre: oh I see. So we need this check in order for the type checker to know that these values are not null and thus match the function definition... is that it?

app/src/hooks/useTransfer.tsx
noahjoeris: I love the Option type in Rust. But in my opinion adding it to TS feels a bit like forcing it artifically into the language. It also makes it harder for other developers to grasp what happens because using undefined is so common in JS/TS. And exposing an api with option param will confuse everyone using it ahaha 😄

As I said I like the option type and would be open to explore if you really want to try it out.

app/src/components/Transfer.tsx
noahjoeris: Correct 💯

app/src/hooks/useTransfer.tsx
NunoAlexandre: I don't think it's that confusing tbh and wouldn't take what's more common as the rule of thumb. If it helps us write better and more resilient code I'd rather do it.

Wdyt?

noahjoeris: In my opinion it has up and downsides. We can try it and throw out if it doesnt work.

Downsides I can think of:
- very uncommon in TS and will confuse other TS devs
- more verbose. Particularly using these variables inside components
- might add a lot of runtime overhead because it is not zero abstraction as in Rust

noahjoeris: Could we maybe include it in another PR then? It must be changed in many places and has nothing to do with the intent of this PR :)

NunoAlexandre: Sure, that's fine. I just want to leave the remark clear that we shouldn't guide ourselves based on what's more common. If it's a hardcore convention that's one thing but when it comes to type safety and handling we should thrive for better approaches. Type safety is always more verbose but that should not be the argument. Now, as I said above, this is not a blocker for me so feel free to move ahead 🌱

NunoAlexandre (approved):

app/src/hooks/useTransfer.tsx
noahjoeris: Yes I'm 100% on your side with this👍

For me it's just that such design decisions are more "risky" and need to be thoroughly considered from all angles.

PR #15, opened 2024-06-03
feat: Add cypress e2e testing

noahjoeris:
This PR integrates Cypress and adds one e2e test


NunoAlexandre:
@noahjoeris do you think we should give something like https://www.meticulous.ai/ a try before pushing for cypress testing? Maybe we could wait before merging this and see what we could get out of a project like that. wdyt?

noahjoeris:
The question is: do we want to do only one or both? Do we entirely want to rely on AI generated tests?

NunoAlexandre:
> The question is: do we want to do only one or both? Do we entirely want to rely on AI generated tests?
> 
> But this tool looks incredible so I definetely would give it a try.

I would say, let;s give the AI alternative a try and if that doesn't suffice us for now we can always go with cypress. Reasoning being we don't have a lot of time to write a a lot of tests manually so if the AI will do it's a win-win

PR #17, opened 2024-06-05
⚡ Feature: switch between evm & substrate wallets and manual address input

noahjoeris:
This PR:
- enables to switch wallet types based on source and destination chain. E.g. switch between evm and substrate.
- enables to toggle between manual receiver address input and wallet connection

Not supported currently:
- two different evm accounts for source and destination
- two different substrate accounts for source and destination

Known issue: 
- Substrate account resets on wallet component removal
    - will be fixed in the future either by storing account in app state or by adopting useInkathon.
- transfer function still doesn’t use the connected substrate wallet address. Will be fixed in a future PR.

image


NunoAlexandre:
@noahjoeris nice, this is cool. 

> transfer function still doesn’t use the connected substrate wallet address. Will be fixed in a future PR.

In my branch I have already created a store and hook for keeping the substrate wallet so once we merge this I can integrate the two parts in my branch 👍

noahjoeris:
Amazing 👍 🚀

app/src/components/Switch.tsx
```diff
@@ -0,0 +1,33 @@
+import { twMerge } from 'tailwind-merge'
+
+interface SwitchProps {
+  /** The current state of the switch, true for on and false for off. */
+  checked: boolean
+  /** Callback invoked with the new state each time the switch is toggled. */
+  onChange: (checked: boolean) => void
```
NunoAlexandre, line 7: is `checked` the value it was at before or the new value?

app/src/components/Transfer.tsx
```diff
@@ -18,7 +24,12 @@ const Transfer: FC = () => {
   const [destinationChain, setDestinationChain] = useState<Chain | null>(null)
   const [token, setToken] = useState<Token | null>(null)
   const [amount, setAmount] = useState<number | null>(null)
-  const [receiverAddress, setReceiverAddress] = useState<string>('')
+  const [manualReceiverAddress, setManualReceiverAddress] = useState<string>('')
+  const [manualAddressInputEnabled, setManualAddressInputEnabled] = useState<boolean>(false)
+  const [walletTypes, setWalletTypes] = useState<{
+    source?: WalletType
```
NunoAlexandre, line 30: Nice 👌

app/src/components/Transfer.tsx
```diff
@@ -37,14 +48,55 @@ const Transfer: FC = () => {
...
+
+  useEffect(() => {
+    if (destinationChain)
+      setWalletTypes(prev => ({ ...prev, destination: chainToWalletType(destinationChain) }))
+  }, [destinationChain])
+
+  const getReceiverAddress = () => {
+    if (manualAddressInputEnabled) return manualReceiverAddress
+    if (walletTypes.destination === WalletType.EVM) return evmAccount.address
+    if (walletTypes.destination === WalletType.SUBSTRATE) return undefined // TODO: Placeholder to get Substrate address from connected wallet. This will be added once substrate connection is fixed.
```
NunoAlexandre, line 65: For me, this function should live next to the definition of WalletType as a `function getAddress` and then we call it as `wallet.destination.getAddress()`.

app/src/components/Transfer.tsx
```diff
@@ -18,7 +24,12 @@ const Transfer: FC = () => {
   const [destinationChain, setDestinationChain] = useState<Chain | null>(null)
   const [token, setToken] = useState<Token | null>(null)
   const [amount, setAmount] = useState<number | null>(null)
-  const [receiverAddress, setReceiverAddress] = useState<string>('')
+  const [manualReceiverAddress, setManualReceiverAddress] = useState<string>('')
+  const [manualAddressInputEnabled, setManualAddressInputEnabled] = useState<boolean>(false)
+  const [walletTypes, setWalletTypes] = useState<{
```
NunoAlexandre, line 29: Let's rename `walletTypes` to just `wallets`

app/src/components/Transfer.tsx
NunoAlexandre: Also here, let's rename `WalletType` to just `Wallet`

app/src/components/Transfer.tsx
```diff
@@ -96,39 +148,60 @@ const Transfer: FC = () => {
...
-              transfer({
-                sourceChain,
-                destinationChain,
-                token,
-                amount,
-                receiverAddress,
-              })
-          }}
+          onClick={handleSubmit}
+          disabled={!isTransferValid()}
```
NunoAlexandre, line 182: I think since the scope of this code is in the `Transfer` component, we don't need to prefix functions with `Transfer`, so this could just be `isValid` imo.

app/src/components/Transfer.tsx
```diff
@@ -96,39 +148,60 @@ const Transfer: FC = () => {
...
+      exit={{ opacity: 0 }}
+    >
+      {walletType === WalletType.EVM && <ConnectEvmWalletButton label="Connect EVM" />}
+      {walletType === WalletType.SUBSTRATE && (
+        <ConnectSubstrateWalletButton label="Connect Substrate" />
+      )}
+    </motion.div>
+  )
+}
+
```
NunoAlexandre, line 206: Can we move this to a separate component?

app/src/models/walletType.ts
```diff
@@ -0,0 +1,18 @@
+import { Chain } from './chain'
+
+export enum WalletType {
+  EVM = 'evm',
+  SUBSTRATE = 'substrate',
+}
```
NunoAlexandre, line 6: I see now how this is defined. I think we would be better off with a different approach, like:

``` typescript
enum WalletType { Substrate, EVM }

type SubstrateWallet = {
  type: WalletType.Substrate
  /// other fields or wrap another standard wallet type
};
type EVMWallet = {
   type: WalletType.EVM
  /// other fields or wrap another standard wallet type
};

// Type that gathers all supported types of wallets
type Wallet =
  | SubstrateWallet
  | NetworkFailedState;
```


With this, the wallet hook / state in `Transfer` would automatically store the wallet data and be ready to support new types of wallet in the future.

app/src/utils/address.ts
```diff
@@ -15,3 +19,28 @@ export const truncateAddress = (str: string, start: number = 4, end: number = 4)
...
+ * @param address - The address string to be validated.
+ * @returns True if the address is a valid Substrate address, false otherwise.
+ */
+export const isValidSubstrateAddress = (address: string): boolean => {
+  try {
+    encodeAddress(isHex(address) ? hexToU8a(address) : decodeAddress(address))
+    return true
+  } catch (error) {
+    return false
+  }
```
NunoAlexandre, line 45: 👍  nice. We will also need a way to verifying that a given address is valid for a given selected chain but we can add that later 👍

app/src/components/Transfer.tsx
NunoAlexandre: Or have a function getAddress that received a wallet type and returns the address as a string

app/src/components/Switch.tsx
noahjoeris: yep, I changed the name to make it clearer 👍

app/src/models/walletType.ts
noahjoeris: Like dat. Thanks! ⭐

app/src/components/Transfer.tsx
noahjoeris: Good point. I think I solved it, please give me your feedback.
- I created a hook useWalletAddress that takes a wallet as input and provides the address.
- for evm it just uses the wagmi useAccount hook
- todo: please integrate the substrate wallet hook/store into it

NunoAlexandre (approved):
Beautiful 🏂

PR #19, opened 2024-06-09
feat: Support bridged transfers Ethereum <> AssetHub

NunoAlexandre:
### Features

- [x] Support for transferring wETH from Ethereum (Sepolia) to (Rococo) AssetHub
- [x] Support for transferring wETH from (Rococo) AssetHub to Ethereum (Sepolia)
- [x] Add `useEthersSigner` hook
- [x] Add `Environment = Mainnet, Testnet` global value that's passed down to the transfer handler to add support for multiple environments
- [x] Add very basic unit tests
- [x] Add store and hook for looking up the connected Substrate wallet

### Changes

- [x] Extend the `Chain` type to include `chainId` and `network`
- [x] Adjust and slim down the `testdata` registry to only contain supported chains
- [x] Refactor the state managment of the selected source & destination chains and the respective connected wallets 
- [x] Deleted some modules / code blocks that became obsolete

### Testing

- [x] Tested transferring wETH from Sepolia to Rococo AssetHub ✅ 
- [x] Tested transferring wETH from Rococo AH to Sepolia ✅ 

### Next (out of scope now)

- Fix `amount` input requiring the right decimals
- Support proper multi-environment registries
- Show ongoing transfer progress


app/src/services/transfer.ts
```diff
@@ -0,0 +1,125 @@
...
+  token: Token,
+  env: Snowbridge.environment.SnowbridgeEnvironment,
+): string => {
+  switch (token.id) {
+    case 'weth':
+      return env.locations[0].erc20tokensReceivable['WETH']
+    default:
+      throw Error('Token not supported or it is not an ERC20 token')
+  }
+}
```
NunoAlexandre, line 47: Temporary: first, looking the token based on the "id" is not very reliable and secondly ideally we don't need to look this info up from another sdk and just have it in our own registry.

app/src/components/Transfer.tsx
```diff
@@ -2,35 +2,33 @@
...
 
 const Transfer: FC = () => {
   // Inputs
   const [sourceChain, setSourceChain] = useState<Chain | null>(null)
   const [destinationChain, setDestinationChain] = useState<Chain | null>(null)
   const [token, setToken] = useState<Token | null>(null)
   const [amount, setAmount] = useState<number | null>(null)
-  const [manualReceiverAddress, setManualReceiverAddress] = useState<string>('')
-  const [manualAddressInputEnabled, setManualAddressInputEnabled] = useState<boolean>(false)
-  const [wallets, setWallets] = useState<{
```
NunoAlexandre, line 29: We were basically storing the network of the selected source/destination chains so I just dropped this and reactively look up that exact data and then call the `useWallet` hook with the right network.

NunoAlexandre (commented):
Some comments for the review

app/src/components/Transfer.tsx
noahjoeris: Nice this seems like a good refactoring!
But the following we still need right?
```
const [manualReceiverAddress, setManualReceiverAddress] = useState<string>('')
const [manualAddressInputEnabled, setManualAddressInputEnabled] = useState<boolean>(false)
```
We could also add a new hook called something like `useAddressInput` to outsource that.
Or just shortened:
`const [manualReceiver, setManualReceiver] = useState<...>({enabled: false, address: ''})`

NunoAlexandre: I didn't remove those lines, I just renamed them to use the "Recipient" terminology to be consistent with the rest of the code and dropped the "address" part due to redundancy. Happy to reconsider!

> const [manualReceiver, setManualReceiver] = useState<...>({enabled: false, address: ''})

I like that, I wanted to do that but honestly was scared to mess it up 😆

app/src/context/snowbridge.tsx
```diff
@@ -0,0 +1,35 @@
+import * as Snowbridge from '@snowbridge/api'
+
+export function getEnvironment(network: string): Snowbridge.environment.SnowbridgeEnvironment {
```
NunoAlexandre, line 3: Not sure if this is the right place for this module

app/src/context/ethers.tsx
```diff
@@ -0,0 +1,22 @@
+import { BrowserProvider, JsonRpcSigner } from 'ethers'
```
NunoAlexandre, line 1: Should probably move this to the hooks directory 👀

NunoAlexandre (commented):
More comments

app/src/hooks/useTransfer.tsx
```diff
@@ -1,60 +1,70 @@
...
-  receiverAddress: string
 }
 
 interface TransferValidationParams {
-  token: Token | null
+  sender: any | null | undefined
   sourceChain: Chain | null
+  token: Token | null
   destinationChain: Chain | null
+  recipient: string | undefined | null
```
noahjoeris, line 24: I usually prefer: `recipient?: string | null` over `recipient: string | undefined | null`.
No blocker but would be good to go for consistency.

app/src/hooks/useTransfer.tsx
```diff
@@ -1,60 +1,70 @@
...
+  token: Token
   destinationChain: Chain
+  recipient: string
   amount: number
-  receiverAddress: string
 }
 
 interface TransferValidationParams {
-  token: Token | null
+  sender: any | null | undefined
```
noahjoeris, line 20: Is there another way? 😅

Maybe:

```typescript
type Sender = JsonRpcSigner | SubstrateAccount
interface TransferValidationParams {
  sender?: Sender | null
  ...
}
```

app/src/hooks/useTransfer.tsx
```diff
@@ -1,60 +1,70 @@
...
+    switch (direction) {
+      case Direction.ToPolkadot:
+        toPolkadot(environment, sender as Signer, token, amount, destinationChain, recipient)
+        break
+      case Direction.ToEthereum:
+        toEthereum(environment, sourceChain, sender as WalletSigner, token, amount, recipient)
+        break
+      default:
+        throw Error('Unsupported flow')
+    }
```
noahjoeris, line 53: I really like dat 💯

app/src/hooks/useWallet.tsx
```diff
@@ -0,0 +1,21 @@
+import useSubstrateWallet from './useSubstrateWallet'
+import useEthersSigner from '@/context/ethers'
+import { Account as SubstrateAccount } from '@/store/substrateWalletStore'
+import { JsonRpcSigner } from 'ethers'
+import { Chain, Network } from '@/models/chain'
+
+const useWallet = (chain: Chain | null): SubstrateAccount | JsonRpcSigner | undefined => {
```
noahjoeris, line 7: can we create a type for this?
`type PlaceholderForAGoodName = SubstrateAccount | JsonRpcSigner`

Makes it more extensible as well I guess.

app/src/services/transfer.ts
```diff
@@ -0,0 +1,125 @@
...
+  token: Token,
+  amount: number,
+  destinationChain: Chain,
+  recipient: string,
+): Promise<void> => {
+  const snowbridgeEnv = getEnvironment(toSnowbridgeNetwork(environment))
+  const context = await getContext(snowbridgeEnv)
+  const tokenContract = getErc20TokenContract(token, snowbridgeEnv)
+
+  await Snowbridge.toPolkadot
```
noahjoeris, line 61: I think this `await` is not correct here.

Usually either `await` or `.then(...)`

app/src/services/transfer.ts
```diff
@@ -0,0 +1,125 @@
...
+      recipient,
+      tokenContract,
+      destinationChain.chainId,
+      BigInt(amount),
+      BigInt(0),
+    )
+    .then(plan => Snowbridge.toPolkadot.send(context, signer, plan))
+    .then(sent => trackToPolkadot(context, sent))
+    .then(res => console.log('Result:', res))
+    .catch(e => console.log('Failed to transfer: ', e))
```
noahjoeris, line 74: The current strategy was to do error handling in a hook that uses the service. To keep the service lean and the hook doing all the notification stuff etc. 

But we can add that later if desired, no problem.

noahjoeris: But I like your clean method chaining. I can see you developed a lot in rust 😄

noahjoeris (commented):
Looks great overall 💯 
Thanks for the effort 🌻

app/src/services/transfer.ts
Victor-Salomon, line 74: We will need to think about a consistent error pattern to handle them properly on the UI. 

Or at least cast them, in a good format 
```
if (error instanceof Error) {
            throw new Error(`Failed to transfer: ${error.message}`);
} else {
            throw new Error(`Failed to transfer: ${error as string}`);
}
```

Victor-Salomon:
Nice job @NunoAlexandre 🔥

app/src/store/environmentStore.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export function toSnowbridgeNetwork(env: Environment): string {
+  switch (env) {
+    case Environment.Mainnet:
+      return 'polkadot_mainnet'
+    case Environment.Testnet:
+      return 'rococo_sepolia'
+  }
+}
```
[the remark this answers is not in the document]
NunoAlexandre: good point, I will move it to the `useEnvironment` hook 👍 

> This just turns our defined enum to something Snowbridge understands right? 

Yes, I tried to encapsulate that in the naming but happy to leave clearer docs!

app/src/hooks/useEnvironment.tsx
```diff
@@ -0,0 +1,10 @@
+import { useEnvironmentStore } from '@/store/environmentStore'
+
+const useEnvironment = () => {
+  const environment = useEnvironmentStore.getState().current
+  const switchTo = useEnvironmentStore.getState().switchTo
```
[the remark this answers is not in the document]
NunoAlexandre: Thanks, Noah! I actually had experienced a lot of issues with the substrate account hook precisely because of that and fixed that one but forgot to update this one. This one didn't cause issues because we are not changing it but it's good to have this ready 🧹

app/src/hooks/useTransfer.tsx
NunoAlexandre: I will go with `sender?: JsonRpcSigner | SubstrateAccount | null` 👍  thanks for noticing this, I set it to any because i it wasn't clear what type the `useEthersSigner` returned 👍

app/src/services/transfer.ts
NunoAlexandre: well spotted 👌  thanks

app/src/services/transfer.ts
NunoAlexandre: thanks for the input, @Victor-Salomon! I will follow up on this in the next work I will do on the ongoing transfers, since we won't be dealing with async errors here so I'd rather only make changes once 👍

app/src/context/snowbridge.tsx
NunoAlexandre: I think this snowbridge module could either stay here or be moved to utils/. The specific `getEnvironment` function was moved to another module (see discussion above).

Let me know your thoughts 👍

app/src/components/SubstrateWalletButton.tsx
```diff
@@ -53,7 +47,7 @@ const SubstrateWalletButton: FC<SubstrateWalletButtonProps> = ({ label = 'Connec
         <button className="btn btn-sm max-w-40 rounded-2xl">{buttonContent}</button>
```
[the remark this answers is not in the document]
NunoAlexandre: @noahjoeris how would you prefer us to apply these change? (given you introduced this part)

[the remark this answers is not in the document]
NunoAlexandre: thanks! I set it to `60vh` so that it scales reasonably well across different screen heights. Could you give it a try with your setup?

NunoAlexandre:
@noahjoeris @Victor-Salomon thanks a ton for the reviews, I think I have addressed everything but please have a look and let me know 🙏

app/src/components/SubstrateWalletButton.tsx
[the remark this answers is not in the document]
noahjoeris: Thanks @Victor-Salomon that looks like a good solution.

noahjoeris:
Last points
- Have you added the max height for the substrate wallet connect modal?
- The deployment currently fails. Can you fix this?

image

noahjoeris:
Perfeeect! 🥇

NunoAlexandre:
> Last point
> 
> * The deployment currently fails. Can you fix this?
> 
> image

yeah sorry I delete file by mistake, fixed already

app/src/components/SubstrateWalletButton.tsx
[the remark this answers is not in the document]
NunoAlexandre: It was done already :)

noahjoeris (approved):

PR #21, opened 2024-06-11
fix: Convert amount to right decimal base

NunoAlexandre:
Currently the app expects the input amount to be already provided with the right decimal base for the selected token.
In this PR, we change that so the user provides a user-facing amount (e.g. 1 WETH) and we convert it to the right decimal base for the selected token.

### Changes

 - Add `decimals` field to the `Token` type
 - Convert the input amount to the right value using `convertAmount`
 - Add unit tests

### DIFF
--- a/app/src/utils/transfer.ts
+++ b/app/src/utils/transfer.ts
@@ -9,8 +9,8 @@ import { Token } from '@/models/token'
  * @param token - the token the amount is relative to
  * @returns the amount in the decimal base of the given token
  */
-export const convertAmount = (input: number | null, token: Token | null): number | null => {
-  if (!input || !token) return null
+export const convertAmount = (input: number | null, token: Token | null): bigint | null => {
+  if (input == null || !token) return null
 
-  return input * 10 ** token.decimals
+  return BigInt(input * 10 ** token.decimals)
 }
--- a/app/src/utils/transfer.ts
+++ b/app/src/utils/transfer.ts
@@ -1,0 +1,16 @@
+import { Token } from '@/models/token'
+
+/**
+ * Convert a user given amount to the correct value that takes the token decimal in consideration
+ *
+ * E.g 1 WETH as input is converted to 1000000000000000000 given that WETH is an 18-decimal token.
+ *
+ * @param input - the input amount
+ * @param token - the token the amount is relative to
+ * @returns the amount in the decimal base of the given token
+ */
+export const convertAmount = (input: number | null, token: Token | null): number | null => {
+  if (!input || !token) return null
+
+  return input * 10 ** token.decimals
+}
--- a/app/src/utils/transfer.ts
+++ b/app/src/utils/transfer.ts
@@ -1,13 +1,13 @@
 import { Token } from '@/models/token'
 
 /**
- * Convert a user given amount to the correct value that takes the token decimal in consideration
+ * Converts a user-specified amount to its corresponding value in the token's decimal base.
  *
- * E.g 1 WETH as input is converted to 1000000000000000000 given that WETH is an 18-decimal token.
+ * For example, if the input is "1 WETH" and given that WETH is an 18 decimals token, the function converts this to 1 * 10^18 = 1000000000000000000 (wei).
  *
- * @param input - the input amount
- * @param token - the token the amount is relative to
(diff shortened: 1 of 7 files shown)


app/src/utils/transfer.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Convert a user given amount to the correct value that takes the token decimal in consideration
+ *
+ * E.g 1 WETH as input is converted to 1000000000000000000 given that WETH is an 18-decimal token.
+ *
+ * @param input - the input amount
+ * @param token - the token the amount is relative to
+ * @returns the amount in the decimal base of the given token
+ */
```
noahjoeris, line 11: I personally would adjust it to make it more clear. But feel free to skip.
```typescript
/**
 * Converts a user-specified amount to its corresponding value in the token's smallest unit (using the token's decimal places).
 *
 * For example, if the input is 1 WETH and WETH has 18 decimal places, the function converts this to 1 * 10^18 = 1000000000000000000 (wei).
 *
 * @param input - The amount specified by the user. For example, 1 WETH.
 * @param token - The token object which includes the number of decimal places.
 * @returns The equivalent amount in the smallest unit of the given token, or null if the input or token is not provided.
 */
```

app/src/utils/transfer.ts
```diff
@@ -0,0 +1,16 @@
...
+export const convertAmount = (input: number | null, token: Token | null): number | null => {
+  if (!input || !token) return null
```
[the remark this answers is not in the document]
NunoAlexandre: that's fine, tho, right? If it's 0 it will convert to 0 anyways 🤷

app/src/utils/transfer.ts
NunoAlexandre: @noahjoeris your suggestion is good but I don't follow what you meant to communicate with this part?

> in the token's smallest unit

app/src/utils/transfer.ts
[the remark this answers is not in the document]
noahjoeris: No if it's 0 it will convert to null

app/src/utils/transfer.ts
noahjoeris: We convert the unit `ETH` to the smallest unit of the Ethereum network called `Wei`. 
That's what we are doing here. Because the network only works with Wei.

noahjoeris: Feel free to change. You wrote `the correct value` which I find a bit vague. Both values are correct values :)

NunoAlexandre:
```suggestion
/**
 * Converts a user-specified amount to its corresponding value in the token's decimal base.
 *
 * For example, if the input is "1 WETH" and given that WETH is an 18 decimals token, the function converts this to 1 * 10^18 = 1000000000000000000 (wei).
 *
 * @param input - The amount specified by the user. For example, 1 WETH.
 * @param token - The token object which includes its decimals property.
 * @returns The amount in with the token's decimal base, or null if the input or token is not provided.
 */
```

Victor-Salomon (approved):
Overall it is all good considering @noahjoeris comments. 
I may have one concern about not using Big Numbers for those amounts.

NunoAlexandre:
> Overall look all good considering @noahjoeris comments. I'm may have just one consern about not using Big Numbers for those kind of amounts?

Yeah that's a good point. I can have a look at that later or on a follow up PR 👍

app/src/utils/transfer.ts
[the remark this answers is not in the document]
NunoAlexandre: That's not what I mean. If the `input` is 0, this returning 0 or `null` is the same in the end given that it's not a valid amount ayways. But I can tweak this, no problem.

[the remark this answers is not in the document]
NunoAlexandre: Done 👍

noahjoeris:
oh yeah important! I overlooked that. We need to use `bigint`.

NunoAlexandre:
@noahjoeris @Victor-Salomon I think I addressed everything, let me have another round 🙏 🚀  thanks!!

noahjoeris (approved):
Looks good 🚀 Thanks.

PR #22, opened 2024-06-11
feat: multi-environment registry

NunoAlexandre:
This is just a first step into defining a better structured for the global
registry we will need for the app to look up the supported chains and
tokens. The registry will either be fully dynamic in that requests will
be made for each need or we will generate a static file every X hours
based on the current state of the different moving pieces (supported
chains and bridges).


app/src/app/api/chains/route.ts
```diff
@@ -1,17 +1,22 @@
...
-const mockChains: Chain[] = testchains
-
 export async function GET(request: NextRequest) {
   const { searchParams } = new URL(request.url)
+  const environmentParam = searchParams.get('environment')
+
+  if (environmentParam == null)
+    return Response.json({ error: "Invalid request: missing 'environment' param" }, { status: 400 })
+  const environment = Environment[environmentParam as keyof typeof Environment]
+
```
noahjoeris, line 13: What if it is a wrong environment?

Maybe we can add:
```typescript
if (!(environmentParam in Environment)) {
    return NextResponse.json({ error: "Invalid request: unknown 'environment' param" }, { status: 400 });
  }
```

app/src/config/registry.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export const REGISTRY = {
+  Mainnet: {
+    chains: [Mainnet.Ethereum, Mainnet.AssetHub],
+    tokens: [Mainnet.WETH],
+  },
+  Testnet: {
+    chains: [Testnet.Sepolia, Testnet.RococoAssetHub],
+    tokens: [Testnet.WETH],
+  },
```
noahjoeris, line 69: I like that registry 🚀

app/src/app/api/chains/route.ts
NunoAlexandre: Good points! I'll look into this later on 👍🏻

NunoAlexandre: @noahjoeris it's fixed, I opted for your suggestion to handle it as a string and lower case it. "environmentParam in Environment" didn't work, neither the `x as keyof typeof _` approach since it tries to match the `x` string as a key of the enum instead of matching it against the string value of each enum key 🤷  I just implemented a function to handle the conversion more explicitly.

PR #30, opened 2024-06-23
🧹 Cleanup:  TokenSelect and ChainSelect

noahjoeris:
This PR:
- removes inter font and ensure dazzed font is applied
- removes unused components
- refactores ChainSelect and TokenSelect
    - Adds hook: useOutsideClick
    - adds and updates types
    - adds components: Dropdown, VerticalDivider
    - renames TokenSelect to TokenAmountSelect


app/src/app/globals.css
```diff
@@ -70,27 +71,28 @@
     --input: 217.2 32.6% 17.5%;
     --ring: 212.7 26.8% 83.9%;
   }
-}
 
-body {
-  font-family: 'Inter', sans-serif;
-}
+  body {
+    font-family: var(--font-dazzed);
```
NunoAlexandre, line 76: Should we also set the default `font-weight: 500` ?

NunoAlexandre (approved):
✨ 🧹

app/src/app/globals.css
Victor-Salomon: For my own curiosity: Is the Dazzed Font tailwind class addition in the body from the Layout file not enough? 

Otherwise, yes, we can add the font weight 500 as we need some other font-weight

Victor-Salomon (approved):
Let's goo 😎

app/src/components/WavesBackground.tsx
```diff
@@ -1,98 +0,0 @@
-import { ComponentPropsWithoutRef } from 'react'
```
Victor-Salomon, line 1: @noahjoeris, maybe we can keep this file, and clean/rename it as we will need the 'wave background"

Otherwise, let's create a new one with the SVG

noahjoeris: Alright 🚀

app/src/app/globals.css
noahjoeris: Yep true we can do both in tailwind. I changed it. 🎸

PR #33, opened 2024-06-26
💫 feat: Add form validation and more

noahjoeris:
This PR

- Adds loading button state with spinner
- Adds clean turtle notifications
- Adds React Hook Form
- Adds zod schema validation. With a 3s delay to show error on amount input. And instant feedback on manual address input as it will be pasted for sure. Doesn't include validation for max available balance.
- Updates chain model
- Refactors useChains to use static registry
- Updates source and dest if they are the same
- Fixes many issues

Note: Please play around with the validation and see if it works as expected!


app/src/models/schemas.ts
```diff
@@ -0,0 +1,52 @@
...
+  .refine(
+    data =>
+      !data.manualRecipient.enabled ||
+      !data.destinationChain ||
+      isValidAddressOfNetwork(data.manualRecipient.address, data.destinationChain.network),
+    {
+      message: 'Invalid address',
+      path: ['manualRecipient', 'address'],
+    },
+  )
```
Victor-Salomon, line 52: Schema looks great 🔥

tokenAmountSchema could have an error message after the refine check in case the token or amount are null. But honestly, it's not mandatory

app/src/components/NotificationToast.tsx
```diff
@@ -74,8 +80,30 @@ const renderSeverityIcon = (severity?: NotificationSeverity) => {
...
+  }
+}
+
+const getSeverityHeader = (severity: NotificationSeverity) => {
+  switch (severity) {
+    case NotificationSeverity.Info:
+      return 'By the way'
+
+    case NotificationSeverity.Error:
+      return 'Oops!'
```
Victor-Salomon, line 97: "Oops! Something went wrong"?

I know we need to handle properly the errors, maybe for now it will be a bit more explicit

Victor-Salomon (approved):

Victor-Salomon:
Great work @noahjoeris 🐢

app/src/models/schemas.ts
noahjoeris: I didn't add that because we agreed to only show errors on invalid input and not on empty input 👍🏻

Victor-Salomon: oh true!

app/src/components/NotificationToast.tsx
noahjoeris: "Oops" was part of the design. It's the notification header. Next to it should be an error message.

app/src/components/Button.tsx
```diff
@@ -27,6 +28,12 @@ const paddingX: Record<ButtonSize, string> = {
   lg: 'px-5',
 }
 
+const spinnerSize: Record<ButtonSize, number> = {
+  sm: 24,
+  md: 24,
```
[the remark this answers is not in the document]
noahjoeris: yes I double checked. You can even see it.
image

PR #35, opened 2024-06-27
feat: Fees preview 

NunoAlexandre:
Working nicely! But I am leaving the look up of the value in dollars to a follow up, we don't need this for the demo.

Also leaving a ticket in our board to handle the tech debt here, namely moving the `useEffect` and the state to a hook maybe 👍


app/src/config/registry.ts
```diff
@@ -27,6 +27,22 @@ export module Mainnet {
     logoURI: 'https://ucarecdn.com/c01c9021-a497-41b5-8597-9ab4e71440c1/wrapped-eth.png',
     decimals: 18,
   }
+
+  export const DOT: Token = {
+    id: 'dot',
+    name: 'Polkadot',
+    logoURI: '',
```
noahjoeris, line 34: logoURI not important for now?

app/src/hooks/useTransfer.tsx
```diff
@@ -220,6 +220,17 @@ const useTransfer = () => {
...
+    sender,
+    sourceChain,
+    token,
+    destinationChain,
+    recipient,
+    amount,
+  }: TransferValidationParams) => {
+    return !!(sender && sourceChain && token && destinationChain && recipient && amount)
+  }
+
```
noahjoeris, line 233: I actually deleted this because we dont need it anymore due to the zod schema.

noahjoeris (approved):
LGTM 👍

app/src/config/registry.ts
NunoAlexandre: no, we won't show the logo of this token since it's not transferrable for now, it's just needed for calculating the fees

Victor-Salomon (approved):

app/src/config/registry.ts
NunoAlexandre: Fixed anyways

app/src/hooks/useTransfer.tsx
NunoAlexandre: Fixed 👍

app/src/components/Transfer.tsx
```diff
@@ -117,9 +124,59 @@ const Transfer: FC = () => {
...
+  useEffect(() => {
+    const fetchFees = async () => {
+      console.log('fetchFees')
+      if (!isValid) {
+        console.log('not valid')
+        setFees(null)
+        return
+      }
+
+      if (!sourceChain || !destinationChain || !tokenAmount || !tokenAmount.token) return
```
noahjoeris, line 141: I think this is redundant. IsValid includes the 2nd if.

noahjoeris (approved):
🆒

app/src/components/Transfer.tsx
NunoAlexandre: It does but I need to have the compiler know that those values are not null below 🙃

PR #37, opened 2024-06-27
registry: Extend mirroring current Snowbridge API

NunoAlexandre:
❗  We should double check all the data added here


app/src/services/transfer.ts
```diff
@@ -40,8 +40,8 @@ export const getErc20TokenContract = (
   env: Snowbridge.environment.SnowbridgeEnvironment,
 ): string => {
   switch (token.id) {
-    case 'weth':
-      return env.locations[0].erc20tokensReceivable[0].address
+    case 'weth' | 'veth':
+      return env.locations[0].erc20tokensReceivable[token.id].address
```
noahjoeris, line 44: Here seems to be an error.
```suggestion
    case 'weth':
    case 'veth':
      return env.locations[0].erc20tokensReceivable[NUMBER].address // TODO: this now takes a number and not a string. So we need to change it
```

noahjoeris: So for WETH it should be index = 0

NunoAlexandre: I have a better fix for this 👍

noahjoeris (approved):
🚀

PR #42, opened 2024-07-03
feat: Lookup ENS name for the connected wallet

NunoAlexandre:
Happy to receive suggestions for a different name rather than "display" 👀 

## To Do
- [ ] Support the name lookup for Polkadot addresses


app/src/components/completed/TransactionDialog.tsx
```diff
@@ -28,14 +28,27 @@ import {
...
+  const [senderDisplay, setSenderDisplay] = useState('')
+  const [recipientDisplay, setRecipientDisplay] = useState('')
+  useEffect(() => {
+    const fetchName = async (address: string, network: Network, setter: (x: string) => void) => {
+      setter((await lookupName(network, address)) ?? truncateAddress(address, 4, 4))
+    }
+
+    fetchName(tx.sender, tx.sourceChain.network, setSenderDisplay)
+    fetchName(tx.recipient, tx.destChain.network, setRecipientDisplay)
   })
```
noahjoeris, line 42: Can you use a hook? Would cleanup a lot and remove redundancy :)
```typescript
import { Network } from '@/models/chain'
import { lookupName } from '@/utils/transfer'
import { useCallback, useEffect, useState } from 'react'

const useLookupName = (network: Network, address: string) => {
  const [display, setDisplay] = useState<string | null>()

  const fetchName = useCallback(async () => {
    try {
      const name = await lookupName(network, address)
      setDisplay(name)
    } catch (e) {
      console.error(e)
    }
  }, [network, address])

  useEffect(() => {
    fetchName()
  }, [network, address])

  return display
}

export default useLookupName

File: app/src/utils/transfer.ts [hunk 2]
```diff
@@ -43,3 +45,21 @@ export const formatDate = (date: string | Date): string => {
...
+        // we are happy to fail silently here for now
+        return null
+      }
+    }
+    case Network.Polkadot: {
+      //todo(nuno)
+      return null
+    }
+  }
+}
```
noahjoeris, line 65: Can we move error handling to the hook `useLookupName`?
```suggestion
export async function lookupName(network: Network, address: string): Promise<string | null> {
  switch (network) {
    case Network.Ethereum: {
      const provider = new ethers.CloudflareProvider()
      return provider.lookupAddress(address)
    }
    case Network.Polkadot: {
      //todo(nuno)
      return null
    }
  }
}
```

noahjoeris (changes requested):
🔥 🔥 🔥

app/src/components/completed/TransactionDialog.tsx
NunoAlexandre: Yeah you are right :) thanks, will do!

Victor-Salomon (approved):

PR #46, opened 2024-07-09
🌟 fix: validation issues

noahjoeris:
- fixes showing a wrong validation state in edge cases
- fixes undesired defocus of input elements


Victor-Salomon (approved):
I already spent some time looking at the code and it LGTM 🔥

PR #47, opened 2024-07-15
fonts: Add Dazzed Bold

NunoAlexandre:


app/src/components/fonts/fonts.ts
```diff
@@ -1,7 +1,15 @@
...
+    {
+      path: './dazzed.woff2',
+      weight: '400',
+    },
+    {
+      path: './dazzed-bold.woff2',
+      weight: '700',
+    },
+  ],
   variable: '--font-dazzed',
```
NunoAlexandre, line 14: I found this different setup to support multiple font weights and it works perfectly. We just have to tag `font-bold` where needed and voila.

What was the `display: swap` for tho? 🤔

NunoAlexandre: @noahjoeris @Victor-Salomon ☝️

noahjoeris (approved):
Beautiful ❤️

PR #49, opened 2024-07-15
refacto ens fetch on transfer component

Victor-Salomon:
This PR contains 
- an improvement of the **useLookupName** to support initialization with **null** values. 
- a small refactoring of fetch ens names from the **Chain Select** fetching component to rely on the useLookupName hook. It should **handle address & network changes to prevent rendering issues. 🤞**

**However:** I have not been able to reproduce the known issue from the mainnet. This PR needs testing on Mainnet as I do not own any ENS for my addresses on Eth mainnet. Tests have been made on Sepolia using a different provider. 
In case the issue is still here, we should consider updating the provider.

### DIFF
--- a/app/src/components/ChainSelect.tsx
+++ b/app/src/components/ChainSelect.tsx
@@ -44,10 +44,7 @@ const ChainSelect = forwardRef<HTMLDivElement, ChainSelectProps>(
     const triggerRef = useRef<HTMLDivElement>(null)
     const dropdownRef = useRef<HTMLDivElement>(null)
     useOutsideClick(triggerRef, dropdownRef, () => setIsOpen(false))
-    const addressLookup = useLookupName(
-      value ? value.network : null,
-      walletAddress ? walletAddress.toLowerCase() : null,
-    )
+    const addressLookup = useLookupName(value?.network, walletAddress?.toLowerCase())
 
     const addressPlaceholder = walletAddress ? truncateAddress(walletAddress, 4, 4) : ''
     const accountName = addressLookup ? addressLookup : addressPlaceholder
--- a/app/src/components/ChainSelect.tsx
+++ b/app/src/components/ChainSelect.tsx
@@ -1,17 +1,19 @@
 'use client'
-import { useOutsideClick } from '@/hooks/useOutsideClick'
+import { forwardRef, useRef, useState } from 'react'
+import Image from 'next/image'
+import { twMerge } from 'tailwind-merge'
+
 import { Chain } from '@/models/chain'
 import { ManualRecipient, SelectProps } from '@/models/select'
-import Image from 'next/image'
-import { forwardRef, useEffect, useRef, useState } from 'react'
-import { twMerge } from 'tailwind-merge'
+import useLookupName from '@/hooks/useLookupName'
+import { useOutsideClick } from '@/hooks/useOutsideClick'
+import { truncateAddress } from '@/utils/address'
+
 import Dropdown from './Dropdown'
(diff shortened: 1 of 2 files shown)


app/src/hooks/useLookupName.tsx
```diff
@@ -1,17 +1,24 @@
+import { useCallback, useEffect, useState } from 'react'
+
 import { Network } from '@/models/chain'
 import { lookupName } from '@/utils/transfer'
-import { useCallback, useEffect, useState } from 'react'
 
-const useLookupName = (network: Network, address: string) => {
-  const [accountName, setAccountName] = useState<string | null>()
+const useLookupName = (network: Network | null, address: string | null) => {
```
noahjoeris, line 6: Can we also do this?
```suggestion
const useLookupName = (network?: Network, address?: string) => {
```

app/src/components/ChainSelect.tsx
```diff
@@ -39,12 +41,16 @@ const ChainSelect = forwardRef<HTMLDivElement, ChainSelectProps>(
...
-    const [accountName, setAccountName] = useState('')
-
     const triggerRef = useRef<HTMLDivElement>(null)
     const dropdownRef = useRef<HTMLDivElement>(null)
-
     useOutsideClick(triggerRef, dropdownRef, () => setIsOpen(false))
+    const addressLookup = useLookupName(
+      value ? value.network : null,
+      walletAddress ? walletAddress.toLowerCase() : null,
+    )
```
noahjoeris, line 50: Can we do this? 
```suggestion
    const addressLookup = useLookupName(value?.network, walletAddress?.toLowerCase())
```

noahjoeris (approved):

app/src/components/ChainSelect.tsx
Victor-Salomon: Sure! 
I did this initially, but because both types from the parent component were `something | null` I wanted to avoid adding undefined to the returned type and ended up providing this version. But the code will be cleaner.

app/src/hooks/useLookupName.tsx
```diff
@@ -1,17 +1,24 @@
...
+    }
+
     try {
       const name = await lookupName(network, address)
       setAccountName(name)
     } catch (e) {
       // Do not throw an error here
-      console.error(e)
+      console.error('Address lookup error:', e)
+      setAccountName(null)
```
NunoAlexandre, line 21: Suppose this had run once and fetched the name correctly and not it's running again but it failed for whatever reason and it set the account name to null... why set it to null? I'd rather keep the last name fetched, no? ENS don't change often at all anyways

NunoAlexandre (commented):
Just a comment 👀

app/src/hooks/useLookupName.tsx
Victor-Salomon: I do agree in a way. I wanted to prevent rendering the ENS in case of an issue. The worst-case scenario is in case of any issue it displays the address. 

But for now, I can remove it for now and see if we need to add it again later. 

There is currently an issue with the app: If you select the ETH as the source chain and input your address and then also select the ETH chain as the dest chain, you should keep seeing the first input address. I fixed it with the first check before the try/catch.

NunoAlexandre: Ah I see. But that's an issue with using a singleton for something that should have two instances (one for `Source` and for `Destination`)

Victor-Salomon: Agreed :)
I've included your comments to my PR. 
I'm merging it

PR #56, opened 2024-07-17
feat: footer

noahjoeris:
- move manual address input alert + add animation
- add footer


NunoAlexandre:
@noahjoeris I'm closing this as we won't be following this approach. Thanks a lot anyways 💛

PR #58, opened 2024-07-18
ad-hoc: Show current balance for selected token

NunoAlexandre:
Show current balance for selected token And tweak input styling more according to the figma specs.

NOTE: This is a temporary design solution until Brandon designs it.

### How it looks

A. No selected token
Screenshot 2024-07-18 at 12 18 23

B. Selected token with balance
Screenshot 2024-07-18 at 12 13 09

C. Selected token without balance
Screenshot 2024-07-18 at 12 18 12

### DIFF
--- a/app/src/components/Transfer.tsx
+++ b/app/src/components/Transfer.tsx
@@ -41,9 +41,7 @@ const Transfer: FC = () => {
     balanceData,
   } = useTransferForm()
 
-  const amountPlaceholder = !isBalanceAvailable ? "Amount" : balanceData?.value == BigInt(0) ? "No balance :(" : `${Number(balanceData?.formatted).toFixed(3).toString() + ' ' + tokenAmount?.token?.symbol} `
-
-  console.log("Transfer - balance is ", balanceData)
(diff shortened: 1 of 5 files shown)


app/src/components/Transfer.tsx
```diff
@@ -38,8 +38,11 @@ const Transfer: FC = () => {
     tokenAmountError,
     manualRecipientError,
     isBalanceAvailable,
+    balanceData,
   } = useTransferForm()
 
+  const amountPlaceholder = tokenAmount?.token == null ? "Amount" : balanceData?.value == BigInt(0) ? "No balance :(" : `${Number(balanceData?.formatted).toFixed(3).toString() + ' ' + tokenAmount?.token?.symbol}`
```
Victor-Salomon, line 44: I think it should also work like this btw the backticks but anyway :)

`${Number(balanceData?.formatted).toFixed(3).toString()}  ${tokenAmount?.token?.symbol}`

Victor-Salomon (approved):
Looks all good to me 😎

brandonhaslegs:
This looks good to me. I think you should remove the `:(` though since it's not necessarily a sad thing. It feels wrong to put it there.

PR #59, opened 2024-07-18
app: Style polish

NunoAlexandre:


app/src/components/Transfer.tsx
```diff
@@ -13,6 +13,7 @@ import Switch from './Switch'
 import TokenAmountSelect from './TokenAmountSelect'
 import WalletButton from './WalletButton'
 import { AnimatePresence, motion } from 'framer-motion'
+import { Image } from '@nextui-org/react'
```
Victor-Salomon, line 16: wrong import :)
It should comes from next js instead of nextui

app/src/components/TurtlesBackground.tsx
```diff
@@ -1,24 +1,36 @@
...
-          className="absolute inset-x-0 bottom-0 h-full w-full lg:h-auto"
-        />
-        <div className="absolute bottom-0 h-1/2 w-full bg-turtle-tertiary lg:hidden" />
+      <div className="turtle-background flex h-[80vh] w-screen flex-col items-center justify-center overflow-hidden bg-[url('/bg.png')] bg-cover bg-bottom">
+        <div className="turtle-dark-overlay flex w-screen flex-col items-center justify-center ">
+        </div>
+      </div>
+      <div>
+        <div className="waves-container">
+          <svg
```
Victor-Salomon, line 12: maybe create a dedicated svg component, unless it will not be used anywhere else :)

Victor-Salomon (approved):
It looks all good. (Be sure to update the Image import) 

Can't wait to test it (with the responsive) after being merged 🔥

It looks so good 😎

app/src/components/Transfer.tsx
NunoAlexandre: Good catch 🔥

PR #60, opened 2024-07-19
Usd tokens value

Victor-Salomon:
This PR contains: 

- **Fetch the tokens' USD value** from Coingecko. 
- Fees fetch either ETH or DOT USD value. 
- Tokens bridged fetch ERC20 USD value from contract address. 

Again already past completed transactions do not have the USD values registered.

_Random improvement_: Completed transaction sorted by latest to oldest.


app/src/config/registry.ts
```diff
@@ -53,7 +53,7 @@ export module Mainnet {
     symbol: 'SHIB',
     logoURI: 'https://s2.coinmarketcap.com/static/img/coins/64x64/5994.png',
     decimals: 18,
-    address: '0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599',
+    address: '0x95ad61b0a150d79219dcf64e1e6cc01f0b64c4ce',
```
NunoAlexandre, line 56: Why the change? 👀  was it wrong or as this accidental?

app/src/utils/transfer.ts
```diff
@@ -63,3 +65,50 @@ export async function lookupName(network: Network, address: string): Promise<str
...
+    }
+    case Network.Polkadot: {
+      if (result.success?.assetHub && 'blockNumber' in result.success.assetHub)
+        return `${removeURLSlash(explorersUrls.subscan_assethub)}/block/${result.success.assetHub.blockNumber.toString()}`
+      return `${removeURLSlash(explorersUrls.subscan_assethub)}/account/${sender}`
+    }
+    default:
+      console.log(`Unsupported network: ${network}`)
+  }
+}
```
NunoAlexandre, line 114: nice 👌

app/src/models/transfer.ts
```diff
@@ -40,19 +41,20 @@ export type TransferResult = TxStatus.Succeeded | TxStatus.Failed
...
   amount: string
-  amountValue?: number
   fees: Fees
   minTokenRecieved?: string
   minTokenRecievedValue?: number
   sender: string
   recipient: string
   date: Date
+  explorerLink?: string
+  hashes?: string[]
```
NunoAlexandre, line 56: where are `hashes` used?

app/src/components/FeesPreview.tsx
```diff
@@ -36,8 +36,9 @@ const FeesPreview: FC<FeesPreviewProps> = ({ loading, fees, hidden }) => {
               <div className="text-turtle-foreground">
                 {feeToHuman(fees)} {fees.token.symbol}
               </div>
-              {/* todo(nuno) */}
-              {/* <div className="text-turtle-level3">${state.fees.inDollars}</div> */}
+              {fees.inDollars > 0 && (
+                <div className="text-turtle-level3">${fees.inDollars.toFixed(10)}</div>
```
NunoAlexandre, line 40: Why `toFixed(10)` here? 🤔  Can we make it `3` instead?

```suggestion
                <div className="text-turtle-level3">${fees.inDollars.toFixed(3)}</div>
```

NunoAlexandre (commented):
@Victor-Salomon great work man, this looks good 👌 ❇️ 

Just a few comments and it's good to go!

app/src/config/registry.ts
Victor-Salomon: SHIB was providing the WBTC contract address
TON was providing the PEPE contract address

NunoAlexandre: uff. it's weird how I did this looking at the Snowbridge SDK and Noah verified it. Probably their sdk was wrong too then 🤔

app/src/components/FeesPreview.tsx
Victor-Salomon: toFixed(3) will not display the fees (at least on testnet) as they are 0.0000005.

I will keep it like this for now as it is part of a proper issue/improvement from our board that needs to be addressed (see [here]
(https://www.notion.so/velocitylabs/df1914bfea20494c8a0a32af79fbb448?v=aadc09f0be594111b29113e58ee73b42&p=0ace274b2d374dd58b78adfe1cf5360f&pm=s))

NunoAlexandre: hm but the price in Dollar should not be a decimal value 🤔

Victor-Salomon: What if fees are less than 1$?

app/src/config/registry.ts
Victor-Salomon: No idea... Let me have again a loop to confirm no more contract are wrong :)

app/src/components/FeesPreview.tsx
NunoAlexandre: Sure but it doesn't happen in practise. It's fine if we want to be sure to display enough decimals but the fees involved in the current transfers on mainnet should all be >1$ and if they are not there's probably an error somewhere.

app/src/models/transfer.ts
Victor-Salomon: Nowhere for now. It was part of the initial design I had in mind. Local storage for completed transfer is a tmp solution anyway. I can skip it if you want.

NunoAlexandre: Yeah I don't think we will use `hashes` so I'd rather remove it

app/src/config/registry.ts
NunoAlexandre: Thanks Victor 🙏

NunoAlexandre:
@Victor-Salomon I think something is not quite right:

Screenshot 2024-07-22 at 10 11 06

See how the value of wETH is always shown as `0` which is incorrect 👀 
Also, we should remove the "Min received" since it's not useful for the coming period since it will always be === amount. We can keep it in the types we store but we just don't need to show it in the UI.

NunoAlexandre:
@Victor-Salomon on the comment above, it's probably actually correct given that wETH price on Sepolia is `$0.00006874` today 👍

Victor-Salomon:
@NunoAlexandre, about the 0.000$. The value is 0. Sepolia contract address can't be fetched on API like coingecko or coinmarketcap as they do not support this testnet contract.

NunoAlexandre (approved):
💵 ✨ 🏄‍♂️  Nice work, Victor!!

PR #66, opened 2024-07-26
feat: cypress e2e tests

noahjoeris:
This PR
- adds some e2e tests. Including filling out the complete form
- allows to connect PJS wallet to the test. Ethereum is not supported yet.
- allows env variables to be set: seed phrase, account address, and manual recipient address
- fixes some minor issues


app/cypress/e2e/wallet.cy.ts
```diff
@@ -0,0 +1,149 @@
...
+      cy.wrap(txRequests.length).should('eq', 1)
+      cy.wrap(txRequests[0].payload.address).should('eq', Alice.address)
+      cy.approveTx(txRequests[0].id)
+      // the tx hash will be present, but the chain will reject the
+      // tx eventually
+      cy.get('#tx-hash').should('not.be.empty')
+      cy.get('#tx-events', { timeout: 10000 }).should('contain', 'system.ExtrinsicFailed')
+    })
+  }) */
+})
```
Victor-Salomon, line 149: If this is never used, is it worth removing it? (unless we want to keep at legacy/exploration)

Victor-Salomon (approved):
It's a lot of configuration, but helpers look clean!! 

I like the 'data-cy=...' implementation! This is seamless, not intrusive in the base code 😍

app/cypress/e2e/wallet.cy.ts
noahjoeris: I want to keep it for now. Might be needed when making a real transfer in the future.

PR #68, opened 2024-07-29
fix 🔨 Replace large background svg with small webp

noahjoeris:
Would be great if you can visit the PR preview and test that it also works on your screens


Victor-Salomon (approved):
The Vercel Preview looks good both on my phone and on my desktop. 

Do you think it could make sens to lazy load the turtlebackground component? I know it should not properly impact the moment it's loaded bc it's displayed on every screens, but I was thinking that perfs might be better anyway?

https://nextjs.org/docs/app/building-your-application/optimizing/lazy-loading

PR #69, opened 2024-07-29
Clean/remove unused files components

Victor-Salomon:


app/tsconfig.json
```diff
@@ -9,6 +9,8 @@
     "target": "esnext",
     "module": "esnext",
     "moduleResolution": "bundler",
+    // "noUnusedLocals": true,
+    // "noUnusedParameters": true,
```
noahjoeris, line 13: Do we need that?

noahjoeris (approved):

app/tsconfig.json
Victor-Salomon: I think so, it's commented now because I wanted to get your insight first. 

It prevents the build from succeeding when some files have unused variables or imports. 

What do you think?

noahjoeris: ah cool. I just added eslint rules to warn unused imports haha

PR #70, opened 2024-07-30
Perf/dynamic imports

Victor-Salomon:
This PR contains a refacto of the homepage component with: 
- Dynamic import 
- Renaming of variables & switch from boolean to enum type


app/src/components/HomeComponentSelect.tsx
```diff
@@ -1,37 +1,43 @@
...
+
+const TransferHistory = dynamic(() => import('@/components/completed/TransactionHistory'), {
+  loading: () => <TransactionLoaderSkeleton />,
+})
 
 export const HomeComponentSelect = () => {
   const { completedTransfers } = useCompletedTransfers()
-  const [isNewTransaction, setIsNewTransaction] = useState<boolean>(true)
-  const isCompletedTransactions = !!completedTransfers && completedTransfers.length > 0
+  const [newTransferInit, setNewTransferInit] = useState<TransferTabOptions>(TransferTab.New)
```
noahjoeris, line 19: Wow love it! :star:

app/src/components/completed/TransactionLoaderSkeleton.tsx
```diff
@@ -0,0 +1,59 @@
...
+            <Skeleton className="h-5 w-20" />
+          </div>
+          <ArrowRight className="h-3 w-3" {...{ fill: colors['turtle-level2'] }} />
+          <div className="flex items-center gap-x-2">
+            <Skeleton className="h-4 w-4 rounded-full border-turtle-level3" />
+            <Skeleton className="h-5 w-20" />
+          </div>
+        </div>
+      </div>
+    </div>
```
noahjoeris, line 38: I always find it hard to read code with so many div wrappers. I don't have a component in my head when I see this.
Ideas to improve it: Simply add doc comments. Or break it down to sub components/functions with clear naming such as header etc. 
Inside the same file.

No blocker. Just ideas. You can keep it if you want.

noahjoeris (approved):
Great

app/src/components/completed/TransactionLoaderSkeleton.tsx
Victor-Salomon: This file contains a placeholder without any proper data but more of a shape because dynamic import can involve a short time to load the component. 

But you are right I will update it a bit for clarity :)

Victor-Salomon: Done

PR #72, opened 2024-08-01
cleanup: useTransferForm and eslint warns

noahjoeris:
This PR
- adds some eslint rules: unused imports and variables, namespace imports, and more typescript rules
- adapt code to new eslint rules
- create useSnowbridgeContext hook


app/src/components/ui/button.tsx
```diff
@@ -1,6 +1,6 @@
-import * as React from 'react'
 import { Slot } from '@radix-ui/react-slot'
 import { cva, type VariantProps } from 'class-variance-authority'
+import React from 'react'
```
Victor-Salomon, line 3:
```
import { ButtonHTMLAttributes, forwardRef } from 'react'

...  extends ButtonHTMLAttributes<HTMLButtonElement>,...
... const Button = forwardRef<HTMLButtonElement, ButtonProps>(...
```

Might be better no?

app/src/components/ui/dialog.tsx
```diff
@@ -1,17 +1,13 @@
 'use client'
 
-import * as React from 'react'
-import * as DialogPrimitive from '@radix-ui/react-dialog'
+import DialogPrimitive from '@radix-ui/react-dialog'
 import { X } from 'lucide-react'
-
+import React from 'react'
```
Victor-Salomon, line 5: may be the same here if you update after my previous comment. 
But according to Shadcn opting for the best practices... I've some doubts about if it needs to be changed or no...

app/src/config/registry.ts
```diff
@@ -112,7 +113,7 @@ export module Mainnet {
 }
 
 /* Testnet :: Rococo - Sepolia */
-export module Testnet {
+export namespace Testnet {
   /* Mainnet :: Polkadot - Ethereum */
```
Victor-Salomon, line 117: Was not aware of this :)

app/src/components/ui/separator.tsx
```diff
@@ -1,7 +1,7 @@
 'use client'
 
-import * as React from 'react'
-import * as SeparatorPrimitive from '@radix-ui/react-separator'
+import SeparatorPrimitive from '@radix-ui/react-separator'
+import React from 'react'
```
Victor-Salomon, line 4: and the same here.

app/src/hooks/useManualRecipientValidation.tsx
```diff
@@ -0,0 +1,34 @@
...
+  }, [destinationChain, manualRecipient.address, manualRecipient.enabled])
+
+  useEffect(() => {
+    validate()
+  }, [manualRecipient.address, destinationChain, manualRecipient.enabled, validate])
+
+  return { error }
+}
+
+export default useManualRecipientValidation
```
Victor-Salomon, line 34: I can't find where is it used?

app/src/hooks/useSnowbridgeContext.tsx
```diff
@@ -0,0 +1,23 @@
+import { getContext, getEnvironment } from '@/context/snowbridge'
+import useEnvironment from '@/hooks/useEnvironment'
+import { Context } from '@snowbridge/api'
+import { useEffect, useState } from 'react'
+
+const useSnowbridgeContext = (): Context | undefined => {
```
Victor-Salomon, line 6: Don't you think we could implement it in the useEnvironement Hook?

app/src/components/ui/button.tsx
noahjoeris: nextjs uses tree shaking to reduce this on build

app/src/config/registry.ts
noahjoeris: yeah I added eslint rules and it complained that you do not use module anymore. But actually it also gives warnings to not use namespaces as well. But for now lets keep it like this

app/src/components/ui/separator.tsx
noahjoeris: lets keep this as well as is

app/src/hooks/useManualRecipientValidation.tsx
noahjoeris: Have not implemented this yet. Lets do that after we fixed more pressing issues

app/src/hooks/useSnowbridgeContext.tsx
noahjoeris: Then we would couple our environment tightly with snowbridges env. Lets see when we fixed the "multiple contexts" issue

PR #73, opened 2024-08-02
feat: extend registry by usdc, usdt and dai

noahjoeris:
- Please double check that I added it correctly [here](https://github.com/Snowfork/snowbridge/blob/main/web/packages/api/src/environment.ts)


Victor-Salomon:
And coingecko ^^

PR #74, opened 2024-08-02
chore: misc improvements

Victor-Salomon:
This PR: 

- abstract progress bar logic in a hook to avoid parent component re-rendering and abstract logic from UI
- add a proper cleanup function to the current polling (it will soon be updated but I already modified this before Noah started his implementation). 
- increase current polling to 2 mins
- refactor bridge status fetching with Tanstack query to add cache
- refactor useSnowbridgeContext with Tanstack query to add cache and implement it everywhere we used to call the getContext promise. 

This PR contains some major breaking changes. Thanks for your review
Tested by creating a new transfer on the testnet.


app/src/components/OngoingTransferDialog.tsx
```diff
@@ -286,21 +289,21 @@ export const OngoingTransferDialog = ({
...
   addCompletedTransfer: (transfer: CompletedTransfer) => void,
-  explorerLink: string | undefined,
+  abortSignal: AbortSignal,
+  explorerLink?: string,
 ) {
-  const snowbridgeEnv = getEnvironment(transfer.environment)
-  const context = await getContext(snowbridgeEnv)
-  /* eslint-disable no-constant-condition */
-  while (true) {
+  // if (abortSignal.aborted) return
```
NunoAlexandre, line 303:
```suggestion
```

NunoAlexandre: delete commented line

app/src/components/OngoingTransferDialog.tsx
```diff
@@ -169,7 +167,12 @@ export const OngoingTransferDialog = ({
               </p>
             </div>
 
-            <ProgressBar progression={progression} outlinedProgressBar={true} />
+            <ProgressBar
+              transfer={transfer}
+              bridgeStatus={bridgeStatus}
+              direction={direction}
+              outlinedProgressBar={true}
```
NunoAlexandre, line 174: can we make `outlinedProgressBar` implicit to `true` since it's always outlined in our design?

app/src/components/ProgressBar.tsx
```diff
@@ -1,12 +1,25 @@
...
 }
 
-const ProgressBar: FC<ProgressBarProps> = ({ progression, outlinedProgressBar }) => {
+const ProgressBar: FC<ProgressBarProps> = ({
+  transfer,
+  direction,
+  bridgeStatus,
+  outlinedProgressBar,
+}) => {
+  const progression = useOngoingTransferProgression(transfer, direction, bridgeStatus)
```
NunoAlexandre, line 21: "progression" is not the right name here, we should rename this to "Progress" instead. Same goes to the hook naming


```suggestion
  const progress = useOngoingTransferProgression(transfer, direction, bridgeStatus)
```

app/src/components/ProgressBar.tsx
```diff
@@ -1,12 +1,25 @@
...
+
   return (
     <>
       {!outlinedProgressBar ? (
```
NunoAlexandre, line 25: I'm not sure what a "outlined" progress bar is?

app/src/hooks/useEnvironment.tsx
```diff
@@ -1,10 +1,14 @@
...
+import { shouldUseTestnet } from '@/utils/env'
 
 const useEnvironment = () => {
-  const environment = useEnvironmentStore(state => state.current)
-  const switchTo = useEnvironmentStore(state => state.switchTo)
+  // Kept this for when we will allow a user to switch between mainnet and testnet.
 
-  return { environment, switchTo }
+  // const environment = useEnvironmentStore(state => state.current)
+  // const switchTo = useEnvironmentStore(state => state.switchTo)
```
NunoAlexandre, line 8: Can we remove this?

app/src/hooks/useOngoingTransferProgression.tsx
```diff
@@ -0,0 +1,37 @@
+import { useEffect, useRef, useState } from 'react'
+
+import { bridgeProgressionValue } from '@/context/snowbridge'
+import { SnowbridgeStatus } from '@/models/snowbridge'
+import { StoredTransfer } from '@/models/transfer'
+import { Direction } from '@/services/transfer'
+
+const useOngoingTransferProgression = (
```
NunoAlexandre, line 8:
```suggestion

const useOngoingTransferProgress = (
```

app/src/hooks/useOngoingTransferProgression.tsx
```diff
@@ -0,0 +1,37 @@
...
+  direction: Direction,
+  bridgeStatus?: SnowbridgeStatus,
+) => {
+  const [progression, setProgression] = useState<number>(0)
+  const progressionIntervalRef = useRef<NodeJS.Timeout | null>(null)
+
+  const getProgression = () => {
+    const transferProgression = bridgeProgressionValue(transfer.date, direction, bridgeStatus)
+    setProgression(transferProgression)
+    if (transferProgression >= 90 && progressionIntervalRef.current) {
```
NunoAlexandre, line 19: It's not clear to me what happens if the progress >+ above 90% by reading this... do we just stop pooling for updates?

NunoAlexandre (approved):
The changes look good overall but I left a couple of comments 👍  two remarks for future PRs:

- We have to split up PRs like this that contain a lot of different work into separate PRs
- We need to keep in mind that not all transfers will be bridge transfers in the near future so we shouldn't hardcode bridging logic into the core of the app

NunoAlexandre (commented):
Sorry, didn't mean to approve 🙉  Review left above

app/src/components/OngoingTransfer.tsx
```diff
@@ -3,21 +3,24 @@ import { FC } from 'react'
...
 import { colors } from '../../tailwind.config'
-import ProgressBar from './ProgressBar'
 
 const OngoingTransfer: FC<{
   transfer: StoredTransfer
   update: string | null
-  progression: number
-}> = ({ transfer, update, progression }) => {
+  direction: Direction
+  bridgeStatus?: SnowbridgeStatus
```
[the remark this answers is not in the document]
Victor-Salomon: Thanks for this feedback. I will update the naming and improve this

app/src/components/OngoingTransferDialog.tsx
```diff
@@ -95,12 +83,22 @@ export const OngoingTransferDialog = ({
     }
 
     pollUpdate()
-  }, [addCompletedTransfer, direction, removeOngoingTransfer, transfer, explorerLink])
+
+    return () => {
+      abortController.abort()
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [addCompletedTransfer, direction, removeOngoingTransfer, transfer, explorerLink]) // // Does this make sense ? Registered Transfer will never change and direction relies on transfer data...
```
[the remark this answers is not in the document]
Victor-Salomon: Actually, this is a true question 🙂. I left the comment considering that Noah is going to update the logic to remove the current polling system.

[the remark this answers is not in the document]
NunoAlexandre: right, that data shouldn't change (unless if we store the latest status of a transfer in the ongoing transfer) so indeed I guess we could remove this?

app/src/components/ProgressBar.tsx
NunoAlexandre: I got it on Figma 👍

app/src/components/OngoingTransferDialog.tsx
Victor-Salomon: we have two different progress bars: One is displayed on the ongoing transfer card with a purple background and no outline border while the second one is displayed in the transfer dialog with the white and outline background. 

Tailwind does not proceed to the progress update when the width is set as the class and the dialog already handles the overflow and index. So this is the way to make the difference between the two versions.

app/src/components/OngoingTransferDialog.tsx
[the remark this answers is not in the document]
Victor-Salomon: Exactly. I'm sure we could remove them. I will let Noah get rid of this as he started updating the hook/polling function.

app/src/components/ProgressBar.tsx
Victor-Salomon: Sure, I will remove progression from everywhere and replace it by progress

app/src/hooks/useOngoingTransferProgression.tsx
Victor-Salomon: Exactly after the transfer progress reaches 90% we stop updating the value as we do not want to display a 100% progress bar that is not updating.

app/src/components/OngoingTransfer.tsx
```diff
@@ -26,8 +29,14 @@ const OngoingTransfer: FC<{
           {formatOngoingTransferDate(transfer.date)}
         </p>
       </div>
+
       {/* Progress bar */}
-      <ProgressBar progression={progression} outlinedProgressBar={false} />
+      <ProgressBar
```
noahjoeris, line 34: you can remove `{/* Progress bar */}` now that we defined a component

app/src/components/OngoingTransferDialog.tsx
NunoAlexandre: it's all good, I had not realised that the progress bar had that nuance in its design in those two contexts :)

app/src/hooks/useOngoingTransferProgression.tsx
NunoAlexandre: cool :)

app/src/components/ProgressBar.tsx
```diff
@@ -1,32 +1,45 @@
...
+import { Direction } from '@/services/transfer'
 import { cn } from '@/utils/cn'
 
 interface ProgressBarProps {
-  progression: number
+  transfer: StoredTransfer
+  direction: Direction
   outlinedProgressBar: boolean
+  bridgeStatus?: SnowbridgeStatus
 }
```
noahjoeris: It is a very specific progress bar. With these props it can only be used for snowbridge transfers. It has more than the responsibility of a progress bar. 

Usually I like to keep such components generic. So having a prop `progress` makes sense to me. But just my opinion.

app/src/hooks/useOngoingTransferProgress.tsx
```diff
@@ -0,0 +1,37 @@
+import { useEffect, useRef, useState } from 'react'
+
+import { getBridgeProgress } from '@/context/snowbridge'
+import { SnowbridgeStatus } from '@/models/snowbridge'
+import { StoredTransfer } from '@/models/transfer'
+import { Direction } from '@/services/transfer'
+
+const useOngoingTransferProgress = (
```
noahjoeris, line 8: As it seems to be snowbridge specific. Can we add that to the name?

noahjoeris (approved):

app/src/components/ProgressBar.tsx
NunoAlexandre: @noahjoeris I agree

PR #75, opened 2024-08-07
update explorer url

Victor-Salomon:
Update the explorer links. 

Note: Polkadot one can't be tested on testnet as it's still broken.


noahjoeris (approved):
LGTM

PR #77, opened 2024-08-07
fix: scroll number input issue

noahjoeris:
This PR
- fixes the scrolling issue on the number input where it increased/decreased the value
- removes a minor nextjs image warning


app/src/components/TokenAmountSelect.tsx
```diff
@@ -94,6 +97,7 @@ const TokenAmountSelect = forwardRef<HTMLDivElement, TokenAmountSelectProps>(
               <ChevronDown strokeWidth={0.2} className="ml-1" />
               <VerticalDivider />
               <input
+                ref={inputRef}
```
Victor-Salomon, line 100: I've already seen onWheel property in the input to handle blur

` onWheel={ event => event.currentTarget.blur() }`

Or:  
       
`onWheel={e => e.target instanceof HTMLElement && e.target.blur()}`

However, I leave you the final code as you might have tried this one

Victor-Salomon (approved):
It looks good to me. 

Maybe worth trying on of the onWheel property if not done. 

Otherwise, it looks all right

app/src/components/TokenAmountSelect.tsx
noahjoeris: You are a genius 🧠

PR #79, opened 2024-08-07
fix: improve status tracking performance

noahjoeris:
This PR
- removes unnecessary history fetches

### DIFF
--- a/app/src/hooks/useSnowbridgeTransferTracker.tsx
+++ b/app/src/hooks/useSnowbridgeTransferTracker.tsx
@@ -6,7 +6,7 @@ import {
   ToPolkadotTransferResult,
   TransferStatus,
 } from '@snowbridge/api/dist/history'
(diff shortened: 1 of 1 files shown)


app/src/hooks/useSnowbridgeTransferTracker.tsx
```diff
@@ -35,19 +35,22 @@ const useSnowbridgeTransferTracker = () => {
     }
   }, [])
 
+  const ongoingTransfersRef = useRef(ongoingTransfers)
+
```
Victor-Salomon, line 39: I'm not sure to understand the need to store the ongoingTransfers in a ref. 

Why not just check ongoingTransfers.length > 0 to prevent fetching while no transfers are ongoing?

noahjoeris: No this will not work. It is needed to update the variable inside the interval.

PR #80, opened 2024-08-07
feat: Prompt for ERC-20 spend approval

NunoAlexandre:
When the user is transferring a token from Ethereum, we now check whether they need to first sign the ERC-20 spend approval and show a nice call to action component so that the user does that before being able to submit the transfer.

Screenshot 2024-08-07 at 21 16 17

### Caveats

- When the user signs the approval, we wait 10s to re-fetch the allowance but sometimes that's not enough so the `TokenSpendApproval` keeps on showing and only a change in the amount or other transfer input field triggers the fix.


app/src/components/Transfer.tsx
```diff
@@ -40,12 +45,24 @@ const Transfer: FC = () => {
...
+    owner: sourceWallet?.sender?.address,
+  })
   const amountPlaceholder =
     !sourceWallet || tokenAmount?.token == null || !sourceWallet.isConnected || !isBalanceAvailable
       ? 'Amount'
       : balanceData?.value == BigInt(0)
         ? 'No balance'
         : `${Number(balanceData?.formatted).toFixed(3).toString() + ' ' + tokenAmount?.token?.symbol}`
+  const requiresErc20SpendApproval =
+    erc20SpendAllowance !== undefined && erc20SpendAllowance < tokenAmount!.amount!
```
NunoAlexandre, line 65: The logic is here that if `erc20SpendAllowance` is set and if it's below the amount, then it's required to be approved. We know here that `tokenAmount` is set because `erc20SpendAllowance` can not be set without a defined tokenAmount beforehand.

Happy to think of a more explicit and idiomatic approach but logically there's no loophole here so I am not worried about the hard cast.

app/src/hooks/useErc20Allowance.tsx
```diff
@@ -0,0 +1,122 @@
...
+        // Now that the user has signed the approval, we wait 10s before fetching the new set value
+        // and cascade the appropriate UI updates.
+        setTimeout(async () => {
+          await fetchAllowance()
+          setApproving(false)
+          addNotification({
+            message: 'Updated ERC-20 spend allowance',
+            severity: NotificationSeverity.Success,
+          })
+        }, 10000)
```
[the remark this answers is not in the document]
NunoAlexandre: If we can fetch the allowance amount the user signed yes! That's what I was suggesting on Slack :) I'll look into that 👍🏻

app/src/components/Transfer.tsx
Victor-Salomon: we should at least add the `&& sourceWallet?.isConnected` as currently if you disconnect it keeps handling the allowance. 

How about if a user asks for an allowance greater than his balance?

app/src/hooks/useErc20Allowance.tsx
```diff
@@ -0,0 +1,122 @@
...
+const useErc20Allowance = ({ network, tokenAmount, owner, context }: Params) => {
+  const { addNotification } = useNotification()
+
+  const [allowance, setAllowance] = useState<number | undefined>()
+  const [loading, setLoading] = useState<boolean>(false)
+  const [approving, setApproving] = useState<boolean>(false)
+
+  const fetchAllowance = useCallback(async () => {
+    if (network !== Network.Ethereum) {
+      console.log('will set to undefined:', network)
```
Victor-Salomon, line 31: can we remove the log?

app/src/hooks/useErc20Allowance.tsx
```diff
@@ -0,0 +1,122 @@
...
+      setApproving(true)
+
+      if (
+        !context ||
+        !network ||
+        network !== Network.Ethereum ||
+        !tokenAmount ||
+        !tokenAmount.amount ||
+        tokenAmount.amount <= 0 ||
+        !tokenAmount.token
```
noahjoeris, line 82: Should we create a function `_basicChecks()`? Because they are very verbose and used twice in this hook. Just an idea.

app/src/hooks/useErc20Allowance.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Hook to fetch the ERC20 token spend allowance for a given token and owner account.
+ */
+const useErc20Allowance = ({ network, tokenAmount, owner, context }: Params) => {
+  const { addNotification } = useNotification()
+
+  const [allowance, setAllowance] = useState<number | undefined>()
+  const [loading, setLoading] = useState<boolean>(false)
+  const [approving, setApproving] = useState<boolean>(false)
```
noahjoeris, line 27: Do we really need these two? Isn't loading enough?

noahjoeris (changes requested):
Looks good 🚀  Left some comments

app/src/hooks/useErc20Allowance.tsx
NunoAlexandre: No. `loading` is regarding the fetching of the allowance; `approving` is about the act of signing the token spend approval. `loading` is actually not used at the moment

app/src/hooks/useErc20Allowance.tsx
NunoAlexandre: The problem with doing that is that the type system will no longer be aware of those checks, so we have to bang them to assert they are not null or undefined 👀  and the checks are slightly different across the two usages too...

app/src/components/Transfer.tsx
NunoAlexandre: > we should just add the && sourceWallet?.isConnected as currently if you disconnect it keeps handling the allowance.


Good point! Let me look into it.


> How about if a user asks for an allowance greater than his balance?

It's fine as is, there's no problem with us displaying a propmpt to increase the allowance even in that scenario 👍

NunoAlexandre: @Victor-Salomon, Fixed 👍

app/src/hooks/useErc20Allowance.tsx
```diff
@@ -0,0 +1,112 @@
...
+      try {
+        await toPolkadot
+          .approveTokenSpend(
+            context,
+            signer,
+            tokenAmount!.token!.address,
+            convertAmount(tokenAmount!.amount, tokenAmount!.token),
+          )
+          .then(x => x.wait())
+          .then(_ => fetchAllowance())
```
Victor-Salomon, line 89: this is so clean now 🐢

NunoAlexandre:
@Victor-Salomon @noahjoeris thanks for your reviews 🤟

PR #81, opened 2024-08-08
ux: Small tweaks ✨ 

NunoAlexandre:


app/src/components/TokenSpendApproval.tsx
```diff
@@ -13,7 +13,7 @@ const TokenSpendApproval: FC<Props> = ({ onClick, approving }) => {
     <div
       className={cn(
         'mt-8 rounded-[20px] bg-turtle-secondary-light px-3 py-5',
-        approving ? 'animate-pulse' : '',
+        approving ? 'animate-pulse cursor-progress' : '',
```
noahjoeris, line 16: When I have a loading cursor my first reaction is always the app got stuck.

app/src/components/TokenSpendApproval.tsx
```diff
@@ -25,7 +25,10 @@ const TokenSpendApproval: FC<Props> = ({ onClick, approving }) => {
           <div className="pb-2 text-left text-small text-turtle-foreground">
             We first need your approval to transfer this token from your wallet.
             <button
-              className="ml-1 text-left text-small text-turtle-foreground underline"
+              className={cn(
+                'ml-1 text-left text-small text-turtle-foreground underline',
+                approving ? 'cursor-not-allowed' : '',
```
noahjoeris, line 30: Do you like `not allowed cursors`? 😅 Personally I find them very ugly.

noahjoeris (approved):
Good eyes Nuno 👁️✔️

app/src/components/TokenSpendApproval.tsx
NunoAlexandre: That's unusual I'd say. I would think that if the component is disabled while the cursor is still displaying normally when hoovering it.

app/src/components/TokenSpendApproval.tsx
NunoAlexandre: I think it's fine, don't find them ugly per se. But we can just use `cursor-progress` here too. Do you prefer that?

app/src/components/TokenSpendApproval.tsx
noahjoeris: I am unusual haha 😄 Let's go then

app/src/components/TokenSpendApproval.tsx
noahjoeris: nah then keep the `not allowed cursor`

app/src/components/TokenSpendApproval.tsx
NunoAlexandre: You are special 💘

Victor-Salomon: I do not feel cursors wrong. 
At least it prevents from crazy moves.

Victor-Salomon (approved):
Lets gooo

PR #84, opened 2024-08-09
feat: turtle 404 error page

Victor-Salomon:
Turtle Error page


noahjoeris (commented):
Nice! Few questions:
- Did you switch intentionally from `global-error.tsx` to `error.tsx`? This doesn"t catch layout errors!
- We show `404` error doesn't matter the actual error. But should be fine for now😃

Victor-Salomon:
image


@noahjoeris, I read again the doc. And as far as I understand it and after my tests this morning to trigger it we need both right ? 

(To trigger the page I did `throw new Error('failed')` in the page.tsx and it was not catched so I replace global-error.tsx by error.tsx

But I think we should have both?

noahjoeris:
From my understanding the global-error should catch everything. But if you tested it feel free to add both.

NunoAlexandre:
@Victor-Salomon the 404 page doesn't look right to me in the preview 👀 

Screenshot 2024-08-12 at 09 00 36

Victor-Salomon:
@NunoAlexandre wired, let me check again.   

As it was more looking like this to me: 

Capture d’écran 2024-08-09 à 12 00 19

NunoAlexandre:
Did you forget to push some changes? Because the content and the text is different.

Victor-Salomon:
I'm checking.. and testing it more. 

But I just replace the global-error.tsx by error.tsx and updated the content. 

Lemme ping you once it looks right again.

NunoAlexandre:
> I'm checking.. and testing it more.
> 
> But I just replace the global-error.tsx by error.tsx and updated the content.
> 
> Lemme ping you once it looks right again.

Merci bien Victor 🤟

app/src/app/not-found.tsx
```diff
@@ -0,0 +1,38 @@
+import Image from 'next/image'
+import Link from 'next/link'
+
+const NotFound = () => {
```
NunoAlexandre, line 4: We could abstract this as a component and pass the title, description, button text and link. That way we could also reuse it for the error pages and other uses like 500 errors

app/src/app/global-error.tsx
```diff
@@ -1,22 +1,58 @@
...
+        </div>
+
+        {/* Content */}
+        <section className="absolute inset-0 z-20 flex flex-col items-center justify-center text-center">
+          <div className="flex flex-col items-center justify-center gap-8 text-white">
+            <h1 className="text-xl font-bold leading-5"> {/* Should be: 500 */}404</h1>
+            <h2 className="text-4xl font-medium leading-[56px] tracking-tighter sm:text-[56px]">
+              {/* Should be: Oops, Turtle is temporary unavailable or eventually Oops, something went wrong.*/}{' '}
+              Oops, this page is missing.
+            </h2>
```
NunoAlexandre, line 46: a 404 shouldn't be the default error page. Is there a way to fetch what the error was in NextJs and build this page accordingly?

NunoAlexandre (approved):

app/src/app/error.tsx
```diff
@@ -1,59 +1,52 @@
 'use client'
-import { captureException } from '@sentry/nextjs'
 import { useEffect } from 'react'
 import Image from 'next/image'
-import Navbar from '@/components/NavBar'
+import Link from 'next/link'
+import { captureException } from '@sentry/nextjs'
 
-interface GlobalErrorProps {
+interface ErrorProps {
```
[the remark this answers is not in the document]
Victor-Salomon: Sure 👌

I'll try to see if any React/Next conventions should be followed.

app/src/app/global-error.tsx
Victor-Salomon: `not-found.tsx` file handles 404 issues. We do not need more content inside this file as the content is clear.

About the `error.tsx` file: 
We do not use many server components for now and no server actions (🥲), but if an error happens it will return a 500 error that will be catched inside this `error.tsx` file. I've written and commented my proposition of 'naming': I thing the title should be: _Oops, something went wrong_. I'm pretty sure I can provide the error detail instead of: `<p>Maybe you clicked on an old link!</p>` and will look into it.

About the `global-error.tsx`: 
Internal/server errors 500 and 404 should be handled by the two previous files. This `global-error.tsx` file will only be triggered for general issues catched within the main/root layout and rendered as the app itself. I would stick to something like: _Turtle is temporarily unavailable_, and be vague about the real issue. Here again, I think I can probably display the error detail, but because it will be displayed once the app has loaded (like a landing page and not during some use of the app), I would not recommend using the error details to avoid unexpected things being displayed. 

Let me know if this is not clear and we can chat about it.

app/src/app/error.tsx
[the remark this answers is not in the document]
Victor-Salomon: It seems that an unformal convention is to suffix props with Props and avoid Prefixing interface in React Function Component with 'I'. 

However, looking at the next js doc, I see every props directly typed in the params. See the screenshot. 
It also avoids having to import FC from React. 

image

I would say the best now is to do this directly in the params unless this is too big. Then let's create a params name type/interface and provide it in the params ({params}:{params}).

app/src/app/not-found.tsx
NunoAlexandre: Ok, thank you for having a look :)

app/src/app/global-error.tsx
NunoAlexandre: It's clear! I like your suggestion of making the text more generic, let's make that change and merge the PR 👍🏻

PR #92, opened 2024-08-13
feat: Improve tx duration estimate

NunoAlexandre:
Transfers from DOT -> ETH take 30 min to 4 hours but currently we just '~30min' for both transfer directions which is confusing and misleading for users.

Here we show `~30 min to 4 hours` for `DOT -> ETH` and `~30 min` for `ETH -> DOT`.


app/src/components/FeesPreview.tsx
```diff
@@ -41,7 +42,7 @@ const FeesPreview: FC<FeesPreviewProps> = ({ loading, fees, hidden }) => {
               )}
             </div>
             <div className="flex items-center">
-              <div className="text-turtle-foreground">~30 mins</div>
+              {estimate && <div className="text-turtle-foreground">{estimate}</div>}
```
noahjoeris, line 45: Do you think we can name it `durationEstimate` to keep things clear

app/src/utils/transfer.ts
```diff
@@ -131,3 +132,24 @@ export function getExplorerLink(transfer: StoredTransfer): string | undefined {
       console.log(`Unsupported network: ${network}`)
   }
 }
+
+//todo(team): query the right sdk to get the appropriate real-time estimate
+export function getEstimate(direction: Direction): string {
```
noahjoeris, line 137: and here maybe `getDurationEstimate`

noahjoeris (approved):
Nice. I left some minor requests. Feel free to ignore.

app/src/components/FeesPreview.tsx
NunoAlexandre: Agreed :) pushed 👍

PR #100, opened 2024-08-15
fix: dropdown going too large

noahjoeris:
- Fixes height
- Adds scroll behavior to dropdowns

### DIFF
--- a/app/pnpm-lock.yaml
+++ b/app/pnpm-lock.yaml
@@ -10,7 +10,7 @@ importers:
     dependencies:
       '@chainsafe/cypress-polkadot-wallet':
         specifier: ^2.2.0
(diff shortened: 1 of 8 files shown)


app/pnpm-lock.yaml
Victor-Salomon, line 1: Did you add any new lib?

app/src/hooks/useEnvironment.tsx
```diff
@@ -1,8 +1,7 @@
 import { Environment } from '@/store/environmentStore'
-import { shouldUseTestnet } from '@/utils/env'
 
 const useEnvironment = () => {
-  const environment = shouldUseTestnet ? Environment.Testnet : Environment.Mainnet
+  const environment = Environment.Mainnet
```
Victor-Salomon, line 4: Did you forget to revert this or should we always work on Mainnet now?

app/pnpm-lock.yaml
noahjoeris: I added one for tailwind scrollbars but removed again. Do I need to revert the lock file?

PR #101, opened 2024-08-15
feat: mobile connection not working note

Victor-Salomon:
This PR adds the WalletNotAccessible component on Talisman Connect on Devices. 
Note: It does not any more rely on Tailwind LG breakpoint but on userAgent native web API.

### DIFF
--- a/app/src/components/WalletNotAccessible.tsx
+++ b/app/src/components/WalletNotAccessible.tsx
@@ -1,0 +1,81 @@
+import { cn } from '@/utils/cn'
+
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogHeader,
+  DialogTitle,
+  DialogTrigger,
+} from '@/components/ui/dialog'
+// import Button from './Button'
+
+export const WalletNotAccessible = ({
+  disabled,
+  isConnected,
+  onClick,
+}: {
+  disabled: boolean
+  isConnected: boolean
+  onClick?: () => void
+}) => {
+  return (
+    <Dialog>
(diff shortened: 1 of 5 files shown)


app/src/components/WalletNotAccessible.tsx
```diff
@@ -0,0 +1,41 @@
+import { cn } from '@/utils/cn'
+import Image from 'next/image'
+
+export const WalletNotAccessible = () => {
+  return (
+    <div className={cn('-mt-8 rounded-[20px] bg-turtle-secondary-light p-3 lg:hidden')}>
```
noahjoeris, line 6: Hmm honestly I do not like the approach of rendering it always and hiding it when the screen size gets larger. Conditional rendering would be better imo.
And is it possible to track based on operating system? Currently you could still see this modal on desktop when you zoom in right?

Victor-Salomon: @noahjoeris, I do agree that this might not be the best option. 
I had investigated before and decided to try this one as the component is temporary and does not contain anything except a bit of CSS & HTML. 

Now you challenged this, I just pushed a conditional rendering option. 
It comes with a drawback: It relies on a regex to check the user's device. 

Another concern: I'm not a huge fan of having this useEffect here. However, this is the 'easiest' solution. We could have fetched the 'header' in a server component, but the only server component accessible is way too far from this component which would include a lot of props drilling 🥲.

noahjoeris (approved):
Amazing @Victor-Salomon  🦸‍♂️

PR #103, opened 2024-08-20
feat: add parachain destination fees

Victor-Salomon:
Add parachain destination fee to Registry + integrate

### DIFF
--- a/app/src/config/registry.ts
+++ b/app/src/config/registry.ts
@@ -35,7 +35,7 @@ export namespace Mainnet {
     logoURI: 'https://s2.coinmarketcap.com/static/img/coins/64x64/22125.png',
     chainId: 3369,
     network: Network.Polkadot,
-    destinationFeeDOT: 500_000_000,
+    destinationFeeDOT: '500000000',
   }
 
   export const WETH: Token = {
--- a/app/src/config/registry.ts
+++ b/app/src/config/registry.ts
@@ -35,6 +35,7 @@ export namespace Mainnet {
     logoURI: 'https://s2.coinmarketcap.com/static/img/coins/64x64/22125.png',
     chainId: 3369,
     network: Network.Polkadot,
+    destinationFeeDOT: 500_000_000,
   }
 
   export const WETH: Token = {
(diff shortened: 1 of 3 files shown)


app/src/config/registry.ts
```diff
@@ -35,6 +35,7 @@ export namespace Mainnet {
     logoURI: 'https://s2.coinmarketcap.com/static/img/coins/64x64/22125.png',
     chainId: 3369,
     network: Network.Polkadot,
+    destinationFeeDOT: 500_000_000,
```
noahjoeris, line 38:
```suggestion
    destinationFeeDOT: 500_000_000n,
```

This way no?

app/src/hooks/useTransfer.tsx
```diff
@@ -158,7 +158,9 @@ const useTransfer = () => {
           token.address,
           destinationChain.chainId,
           amount,
-          500_000_000n,
+          destinationChain.destinationFeeDOT
+            ? BigInt(destinationChain.destinationFeeDOT)
+            : BigInt(0),
```
noahjoeris, line 163:
```suggestion
          destinationChain.destinationFeeDOT || 0n,
```

app/src/models/chain.ts
```diff
@@ -13,6 +13,8 @@ export interface Chain {
   chainId: number
   /** Network the chain belongs to, e.g., Polkadot, Ethereum. */
   network: Network
+  /** Optionnal parameter for parachain only */
```
noahjoeris, line 16:
```suggestion
  /** Optional parameter for parachains only. Used for xcm transfers */
```

noahjoeris (changes requested):
left some comments

app/src/config/registry.ts
Victor-Salomon: I actually did this on purpose. It's gonna be stored on the local storage and it does not handle BN properly. This value should just be used in the useTransfer hook which expects a BN and wield if you try to provide a number. This is why I've decided to go this way

PR #105, opened 2024-08-21
feat: add support for eth to para flow

noahjoeris:
- support eth -> Mythos transfers
- adapt to chain address types
- update registry and change its structure


app/src/hooks/useTransferForm.tsx
```diff
@@ -203,7 +212,7 @@ const useTransferForm = () => {
     const isValidAddress =
       !manualRecipient.enabled ||
       !destinationChain ||
-      isValidAddressOfNetwork(manualRecipient.address, destinationChain.network) ||
+      isValidAddressOfTypes(manualRecipient.address, destinationChain.supportedAddressTypes) ||
```
NunoAlexandre, line 215: `isValidAddressOfTypes` is an odd name. How about `isSupportedAddressType` or `isValidAddressType`

app/src/models/chain.ts
```diff
@@ -11,11 +13,21 @@ export interface Chain {
    * Different networks may have chains with the same chain ID.
    */
   chainId: number
+  /** Chain uids that tokens can be transferred to. */
+  transferableTo: string[] // uid
```
NunoAlexandre, line 17: Regarding the `transferableTo` property 👀 



I don't think this should live here and this property doesn't describe the "chain uids that tokens can be transferred to", because between any chain A and chain B that depends per token.  Semantically it's also not quite there since a chain is not "transferrable" to anywhere. What we want to capture here is whether it's possible to route a transfer from chain A to chain B. That still misses the token variable, which we can only know based on whether chain B has that token registered properly.

My suggestion is to have a dedicated `supportedRoutes` list or map at the high level in the registry that lists all the possible routes. For example:

```
const supportedRoutes = {
  'ethereum': ['assethub', 'mythos', 'bifrost', ...],
  'mythos': 'ethereum',
  'assethub': 'ethereum', 'bifrost' ]
}
```

and then have a function that checks if A to B is possible:

```
function isRouteSupported(a, b) -> boolean {
  return supportedRoutes[a].contains(b)
}
```

NunoAlexandre: Actually, thinking about this a bit more, I think this `supportedRoutes` idea is still missing the target. I'd say that to have one source of truth and reasoning, we should consider a transfer possible simply if `chainA` _and_ `chainB` both contain the target token. That gives us all the information.

wdyt?

NunoAlexandre: of course, that would mean that the destination chain is disabled until the user has selected a token. that might entail a drawback I am not seeing now, as that seems fine. food for thought.

But the point being that if we have first a check whether chain A and chain B are connected, that's not sufficient because they might be connected to transfer, for example, `wETH` but `chainB` might not support `USDC` so that setup is not useful enough for that case.

NunoAlexandre (commented):
Amazing work, @noahjoeris 🌱 You [ridin'](https://www.youtube.com/watch?v=CtwJvgPJ9xw) 🤟 

Just an important thread about the `transferableTo` property 👀

app/src/components/Transfer.tsx
```diff
@@ -12,13 +17,8 @@ import SubstrateWalletModal from './SubstrateWalletModal'
 import { AlertIcon } from './svg/AlertIcon'
 import Switch from './Switch'
 import TokenAmountSelect from './TokenAmountSelect'
-import WalletButton from './WalletButton'
 import TokenSpendApproval from './TokenSpendApproval'
-import useSnowbridgeContext from '@/hooks/useSnowbridgeContext'
```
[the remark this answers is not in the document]
noahjoeris: I simply use Prettier and it formats my imports.
Here are my settings:
```json  
"[typescript]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode",
    "editor.formatOnSave": true
  },
  "[typescriptreact]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode",
    "editor.formatOnSave": true
  },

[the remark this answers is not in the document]
noahjoeris: oh and I found this setting:
```json
"notebook.codeActionsOnSave": {
    "\"source.organizeImports\"": true
}

NunoAlexandre (approved):
Beautiful

PR #106, opened 2024-08-21
fix: Set letter-spacing according to the Figma design spec

### LINKED ISSUE #95: Add correct letter spacing to fonts
Add 0.03em letter-spacing to Large Medium and Small text styles.

image

NunoAlexandre:
Fixes #95

### DIFF
--- a/app/src/app/globals.css
+++ b/app/src/app/globals.css
@@ -36,8 +36,9 @@
 
     --radius: 0.5rem;
 
-    --letter-spacing-xs: -0.04em;
-    --letter-spacing-sm: -0.03em;
+    --letter-spacing-xxl: -0.04em;
+    --letter-spacing-xl: -0.03em;
+    --letter-spacing-default: 0.03em;
   }
 
   .dark {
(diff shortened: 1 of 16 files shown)


app/src/app/globals.css
```diff
@@ -36,8 +36,9 @@
 
     --radius: 0.5rem;
 
-    --letter-spacing-xs: -0.04em;
-    --letter-spacing-sm: -0.03em;
+    --letter-spacing-xxl: -0.04em;
+    --letter-spacing-xl: -0.03em;
```
Victor-Salomon, line 40: Does our app design system follow some specific rules such as every h1 having 0.04em letter spacing and other h2, h3, h4, and h5 with 0.03em?

NunoAlexandre: no. the figma spec just specifies these groups of letter spacing that I defined here with no specific attribute to any specific html tag.

Victor-Salomon (approved):
All right! I've left one question. 
If the answer is yes, maybe we can default letter spacing in our global css for headers like <h1></h1> and <p></p> ?

PR #108, opened 2024-08-22
Feat: swap from to chains

Victor-Salomon:
Implementation of Brandon suggested design to swap from/to chains


app/src/components/Transfer.tsx
```diff
@@ -30,6 +31,7 @@ const Transfer: FC = () => {
     handleSubmit,
     handleSourceChainChange,
     handleDestinationChainChange,
+    handleSwapFromToChains,
```
noahjoeris, line 34: rename to handleSwapChains?

noahjoeris (approved):
Nice 🚀 

In the future we might need to check if we can actually swap the chains or if it's not possible.
And you decided to deselect the token on swap. I think that's fine for now.

app/src/components/SwapFromToChains.tsx
```diff
@@ -0,0 +1,21 @@
+import { motion } from 'framer-motion'
+import { Swap } from './svg/Swap'
+
+export const SwapFromToChains = ({ handleChainChange }: { handleChainChange: () => void }) => {
```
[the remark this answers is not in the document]
NunoAlexandre: Agree with @noahjoeris.

And nice idea passing the `handleChainChange` as a prop 👍 I'd also rename that to `onSwap` or `onChange`

app/src/hooks/useTransferForm.tsx
```diff
@@ -133,9 +133,34 @@ const useTransferForm = () => {
...
+    // Notify swap chains
+    addNotification({
+      severity: NotificationSeverity.Default,
+      message: 'Chains swapped',
+      dismissible: true,
+    })
+    // Reset selected token
+    if (tokenAmount && tokenAmount.token) {
+      resetField('tokenAmount')
+    }
```
NunoAlexandre, line 161: This is quite undesirable from a UX perspective but I understand why we are doing this. Can you leave a `TODO` for us to handle this properly once we support bidirectional transfers? 🙏

NunoAlexandre (commented):
That was fast 🚤 ⚡ 🤩

app/src/hooks/useTransferForm.tsx
```diff
@@ -133,9 +133,34 @@ const useTransferForm = () => {
...
+  const handleSwapFromToChains = useCallback(() => {
+    if (!sourceChain && !destinationChain) return
+    // Swap chains values
+    setValue('sourceChain', destinationChain)
+    setValue('destinationChain', sourceChain)
+    // Notify swap chains
+    addNotification({
+      severity: NotificationSeverity.Default,
+      message: 'Chains swapped',
+      dismissible: true,
```
NunoAlexandre, line 156: Personally I would remove this, I don't like to see notifications popping as a user for something I did intentionally

NunoAlexandre:
Screenshot 2024-08-23 at 10 34 29

@Victor-Salomon I already left the remark about letter-spacing on a comment but just leaving here the side by side between how it looks now and the Figma spec. The font looks bolder on the actual app but I see that the Settings (apart from letter-spacing) are all correct. The same happens for (most) other elements. I wonder if we can tweak some setting to improve the font rendering on browser 👁️

NunoAlexandre:
Setting these css properties to the `body` helps!

``` css
-webkit-font-smoothing: antialiased;
-moz-osx-font-smoothing: grayscale;
font-smooth: never;
```

_Before_
Screenshot 2024-08-23 at 10 46 36

_After_
Screenshot 2024-08-23 at 10 46 47

PR #110, opened 2024-08-23
update chain filters

Victor-Salomon:
Implementation of the chains transfer filters (source chains, dest chains, tokens).

Now implementing isRouteAllowed filter and implementing it in chain swap.

### DIFF
--- a/app/src/utils/filters.ts
+++ b/app/src/utils/filters.ts
@@ -3,8 +3,8 @@ import { Chain } from '@/models/chain'
 import { Token } from '@/models/token'
 import { Environment } from '@/store/environmentStore'
 
-/** Filters all chains by compatibility for the selected destination and token. */
-export const getFilteredSourceChains = (env: Environment): (Chain & { allowed: boolean })[] => {
+/** Filters all chains by available routes. */
+export const getAllowedSourceChains = (env: Environment): (Chain & { allowed: boolean })[] => {
   const routes = REGISTRY[env].routes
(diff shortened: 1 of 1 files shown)


app/src/utils/filters.ts
```diff
@@ -4,33 +4,65 @@ import { Token } from '@/models/token'
...
+    if (r.to === destinationChain.uid && r.tokens.includes(token.id)) {
+      sourceChains.push(r.from)
+      return r
+    }
+  })
+
+  return {
+    sourceChainsData,
+    sourceChains,
+  }
```
noahjoeris, line 25: I think this should be dependent on `env` only. So it should contain all chains that have at least one route available. 
So it can be simplified.

app/src/utils/filters.ts
```diff
@@ -4,33 +4,65 @@ import { Token } from '@/models/token'
...
   env: Environment,
-  _sourceChain: Chain | null,
-  _destinationChain: Chain | null,
+  sourceChain: Chain | null,
+  destinationChain: Chain | null,
 ) => {
-  const tokens = REGISTRY[env].tokens
+  if (!sourceChain || !destinationChain) return
+
+  const tokens: string[] = []
```
noahjoeris, line 58: you have defined the vars as const

Victor-Salomon: yes I could not push my commit with let 
image

PR #112, opened 2024-08-23
feat: allow swap according to selected data

Victor-Salomon:


app/src/hooks/useTransferForm.tsx
```diff
@@ -92,6 +92,18 @@ const useTransferForm = () => {
     (!manualRecipient.enabled || manualRecipient.address.length > 0) &&
     (manualRecipient.enabled || destinationWallet?.isConnected)
 
+  const allowSwap = useCallback(() => {
+    if (!sourceChain || !destinationChain || !tokenAmount) return false
```
noahjoeris, line 96:
```suggestion
    if (!sourceChain || !destinationChain || !tokenAmount?.token) return false
```

this way no?

app/src/components/Transfer.tsx
```diff
@@ -149,7 +150,7 @@ const Transfer: FC = () => {
         />
 
         {/* Switch source and destination chains */}
-        {sourceChain && destinationChain && (
+        {sourceChain && destinationChain && allowSwap() && (
           <AnimatePresence>
             <SwapChains handleChainChange={handleSwapChains} />
           </AnimatePresence>
```
noahjoeris, line 156: On unallowed swap, do we want either
- hide the swap component
- disable the swap component

I like disable more honestly. I do not like this frequent layout restructuring

noahjoeris (commented):
nice. left some points

app/src/hooks/useTransferForm.tsx
```diff
@@ -92,6 +92,18 @@ const useTransferForm = () => {
...
+    return (
+      getAllowedSourceChains(environment).some(
+        sc => sc.allowed && sc.chainId === destinationChain.chainId,
+      ) &&
+      getAllowedDestinationChains(environment, destinationChain, tokenAmount.token).some(
+        dc => dc.allowed && dc.uid === sourceChain.uid,
+      )
```
NunoAlexandre, line 103: It would be cool to move these expressions with the `.some` into separate functions (like (isRoutePossible(environment, source, dest)` so that this function can be more declarative:

``` typescript
return !!sourceChain && !!destinationChain && !!tokenAmount && isRoutePossible(environment, source, dest) && isRoutePossible(environment, dest, source)
```

NunoAlexandre (approved):
Looks good, just one remark to improve the code a bit :)

app/src/hooks/useTransferForm.tsx
Victor-Salomon: I've tried both. This way it works well as the aim is to prevent having tokenAmount completely null. It will work the same. But I can increase the narrowing.

app/src/components/Transfer.tsx
NunoAlexandre: I also don't like to show and hide stuff all the time but I'm wondering how disabling this component will look like since it's all gray already. let's give it a try and see

noahjoeris (approved):

app/src/components/Transfer.tsx
```diff
@@ -149,11 +150,7 @@ const Transfer: FC = () => {
         />
 
         {/* Switch source and destination chains */}
-        {sourceChain && destinationChain && (
-          <AnimatePresence>
-            <SwapChains handleChainChange={handleSwapChains} />
-          </AnimatePresence>
-        )}
+        <SwapChains handleChainChange={handleSwapChains} swapAllowed={allowSwap()} />
```
noahjoeris, line 153: Can we go with `disabled` instead of swapAllowed?

PR #113, opened 2024-08-26
feat: adapt ui to routes

noahjoeris:
This PR
- adds chain and token ordering based on routes
- resets dest and token on source chain change
- fixes issue of token deselect on swap
- adds some cleanups

### DIFF
--- a/app/src/hooks/useTransferForm.tsx
+++ b/app/src/hooks/useTransferForm.tsx
@@ -41,7 +41,6 @@ const useTransferForm = () => {
     handleSubmit,
     setValue,
     reset,
-    resetField,
     formState: { errors, isValid: isValidZodSchema, isValidating },
   } = useForm<FormInputs>({
     resolver: zodResolver(schema),
@@ -133,7 +132,7 @@ const useTransferForm = () => {
     // Swap chains values
     setValue('sourceChain', destinationChain)
     setValue('destinationChain', sourceChain)
-  }, [sourceChain, destinationChain, setValue, tokenAmount])
+  }, [sourceChain, destinationChain, setValue])
 
   const handleManualRecipientChange = useCallback(
     (newValue: ManualRecipient) => setValue('manualRecipient', newValue),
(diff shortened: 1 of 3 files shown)


NunoAlexandre (commented):
We shouldn't reset the token and dest chain when changing the source as long as the route is still supported. 

What the rationale?

noahjoeris:
fixed

app/src/utils/filters.ts
```diff
@@ -62,7 +63,11 @@ export const getAllowedTokens = (
     }
   })
 
-  return tokens
+  return orderByAllowedTag(tokens) as (Token & { allowed: boolean })[]
```
Victor-Salomon, line 66: should we get this (Token & { allowed: boolean }) as a proper reusable type ?

app/src/hooks/useTransferForm.tsx
```diff
@@ -104,38 +100,38 @@ const useTransferForm = () => {
...
+
+      if (!newValue || newValue.uid === sourceChain?.uid) return
+
+      const isSameDestination = destinationChain?.uid === newValue.uid
+
+      if (
+        destinationChain &&
+        tokenAmount &&
+        !isSameDestination &&
+        isRouteAllowed(environment, destinationChain, newValue, tokenAmount)
```
Victor-Salomon, line 113: I will soon PR a better version of this helper. I will ensure that I'm fixing any potential issues/conflicts.

app/src/utils/filters.ts
noahjoeris: We could. But the question is, will we use it again in the future

app/src/hooks/useTransferForm.tsx
noahjoeris: Amazing. Yeah I don't like my way that much.

app/src/utils/filters.ts
Victor-Salomon: You are probably right. Probably not so let's not get annoyed by this anyway

app/src/hooks/useTransferForm.tsx
Victor-Salomon: I was speaking about the `isRouteAllowed` helper only 😅 (I already drafted the PR)

This one largely does what it's supposed to do so we will soon move on 🦾

PR #116, opened 2024-08-27
fix: ⭐ token list update on dest chain switch

noahjoeris:
- Fixes the case where destination is changed but tokens are not updated
- Fixes double clicking the swap button from selecting the text

### DIFF
--- a/app/src/components/SwapFromToChains.tsx
+++ b/app/src/components/SwapFromToChains.tsx
@@ -1,5 +1,5 @@
-import { Swap } from './svg/Swap'
 import { cn } from '@/utils/cn'
+import { Swap } from './svg/Swap'
 
 export const SwapChains = ({
   disabled,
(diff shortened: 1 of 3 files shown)


app/src/components/SwapFromToChains.tsx
```diff
@@ -12,7 +12,7 @@ export const SwapChains = ({
     <div
       onClick={!disabled ? handleChainChange : undefined}
       className={cn(
-        '-my-4 mx-auto flex  items-center justify-center space-x-0.5 p-2 text-turtle-level6',
+        '-my-4 mx-auto flex  select-none items-center justify-center space-x-0.5 p-2 text-turtle-level6',
```
Victor-Salomon, line 15: it works well 👏

Victor-Salomon (approved):
Branch tested locally. 
It works like a charm. 
Nice addition 🐢

A point to notice in the filter.ts folder: 
Should we improve consistency in our types regarding `paramName?: type | null` ?

PR #117, opened 2024-08-28
Fix: prevents swap when transfer is ongoing

Victor-Salomon:
This PR prevent chains swap when a transfer is ongoing and updates CSS accordingly.

### DIFF
--- a/app/src/components/Transfer.tsx
+++ b/app/src/components/Transfer.tsx
@@ -150,7 +150,11 @@ const Transfer: FC = () => {
         />
 
         {/* Switch source and destination chains */}
-        <SwapChains handleChainChange={handleSwapChains} disabled={!allowSwap()} />
+        <SwapChains
+          handleChainChange={handleSwapChains}
+          disabled={!allowSwap()}
+          isTransferOngoing={transferStatus !== 'Idle'}
+        />
 
         {/* Destination Chain */}
         <Controller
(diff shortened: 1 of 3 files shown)


app/src/components/Transfer.tsx
```diff
@@ -150,7 +150,11 @@ const Transfer: FC = () => {
         />
 
         {/* Switch source and destination chains */}
-        <SwapChains handleChainChange={handleSwapChains} disabled={!allowSwap()} />
+        <SwapChains
+          handleChainChange={handleSwapChains}
+          disabled={!allowSwap()}
+          isTransferOngoing={transferStatus !== 'Idle'}
+        />
```
noahjoeris, line 157:
```suggestion
        <SwapChains
          handleChainChange={handleSwapChains}
          disabled={!allowSwap() || transferStatus !== 'Idle'}
        />
```

Victor-Salomon:
I already included the check within the allowSwap, and the handleSwap. The fact is, the swap can be disabled because the chain swap is not allowed (which relies on the allowSwap) or because of an ongoing transfer (which relies on the status itself). I will check again if I can improve.

Victor-Salomon:
@noahjoeris I can't bypass the cursor modification without taking into consideration an ongoing transfer (which differs from the 'disabled status'). When the component is disabled because of an ongoing transfer, the cursor changes from default to not allowed on hover which is wired. 

Either we: 
- keep the hover cursor switch (I don't like this)
- remove the cursor not allowed to default when the component is disabled (no matter if this is because the chains are not swappable or an ongoing transfer)
- validate the PR with this third props

I would go for option 2

noahjoeris:
Alright thanks. You can decide. I'd also go with option 2 or 3

noahjoeris (approved):

PR #119, opened 2024-08-29
fix: add missing validation param

Victor-Salomon:
Fix schema validation that blocked the ETH => Para transfer.


noahjoeris (approved):
missed that one

PR #121, opened 2024-09-05
Feat: AT API transfer tracking

Victor-Salomon:
This PR: 

- refactor the .signAndSend from AT API transfers by handling the transfer in a callback that filters events and data once txHash is included in the block. 

- Handles a parachain to ETH tracking using the new subscan endpoints(_tracking itself + the handlers it impacts_). We do not support the direct para to eth flow yet, so the tracking case should not happen. However, the code is here and can be tested by switching the tracking in a AH to ETH transfer from Snowbridge history approach to XCMSubscan. The subscan file contains a lot of commented code from past research and tests I've made. _Until further notice on the potential para-to-Eth, 2 step process, lets keep it as legacy code since it can be useful. **(No need to review it)**._

- Refactor the hook that add the ongoing transfer to storage to prevent adding twice the same transfer.


app/src/utils/snowbridge.ts
```diff
@@ -97,8 +106,24 @@ export async function getTransferHistory(
...
+    // )
+    // console.log('From AH To Ethereum transfer:', toEthereum.length)
+    // transfers.push(...toEthereum)
+  }
+
+  if (ongoingTransfers.toEthereum.fromParachain.length) {
+    const fromParachainTransfer = await trackXcmTransfer(
+      relaychainScan,
+      ongoingTransfers.toEthereum.fromParachain,
+    )
```
noahjoeris, line 124: Thought: Now that we use also subscan-xcm tracking is it placed correctly here in `util/snowbridge`?

app/src/utils/snowbridge.ts
```diff
@@ -129,23 +152,36 @@ export interface AccountInfo {
...
+      )
         return 'Arriving at Ethereum'
-      if (submitted) return 'Arriving at Bridge Hub'
+      if (
+        'submitted' in transferResult ||
+        !transferResult.destEventIndex ||
+        ('destEventIndex' in transferResult && transferResult.destEventIndex.length === 0)
+      )
+        return 'Arriving at Bridge Hub'
       return 'Pending'
```
noahjoeris, line 185: Is it possible to either add simple comments here or better derived variables? These `if's` are not readable anymore. It's dark magic 🪄 
like
`const xyzHappened = some bool expression`
and then use it in the `if (xyzHappened || ...)`

app/src/utils/snowbridge.ts
```diff
@@ -179,36 +215,20 @@ export function getTransferStatusToPolkadot(transferResult: ToPolkadotTransferRe
...
-    toEthereumOperatingMode: StatusValue
-    overallStatus: StatusValue
-  }
-}
+const getTimestamp = (
+  transferResult: ToEthereumTransferResult | ToPolkadotTransferResult | SubscanXCMTransferResult,
+) =>
+  'info' in transferResult
+    ? transferResult.info.when.getTime()
+    : transferResult.originBlockTimestamp
```
noahjoeris, line 231: Also here we couple xcm tracking with snowbridge:
probably makes sense to decouple it into multiple files. And call it from one generic function.

app/src/hooks/useAssetTransferApi.tsx
```diff
@@ -57,60 +57,106 @@ const useAssetTransferApi = () => {
...
-          destinationChain: destinationChain.name,
-          usdValue: tokenUSDValue,
-          usdFees: fees.inDollars,
-          recipient: recipient,
-          date: date.toISOString(),
+        .signAndSend(account.address, { signer: account.signer as any }, async result => {
+          // verify transaction hash & transfer isn't completed
+          if (!result.txHash) {
+            throw new Error('Transfer failed')
+          }
```
noahjoeris, line 69:
```suggestion
          if (!result.txHash) throw new Error('Transfer failed')
```

app/src/hooks/useAssetTransferApi.tsx
```diff
@@ -57,60 +57,106 @@ const useAssetTransferApi = () => {
...
+                messageId = (data.messageId as any).toString()
+              }
+              if (method === 'ExtrinsicSuccess' && section === 'system') {
+                extrinsicSuccess = true
+              }
+            })
+
+            if (!extrinsicSuccess)
+              throw new Error('Transfer failed. Returned extrinsicSuccess: false')
+            if (!messageHash) throw new Error('Crosschain messageHash missing')
```
noahjoeris, line 101:
```suggestion
            if (!messageHash) throw new Error('Cross chain messageHash missing')
```

noahjoeris (commented):
Great work @Victor-Salomon 🚀 
Left some comments, suggestions and ideas ✨

app/src/utils/subscan.ts
```diff
@@ -0,0 +1,608 @@
...
+
+export const trackXcmTransfer = async (
+  relaychainScan: subscan.SubscanApi,
+  ongoingTransfers: OngoingTransferWithDirection[],
+) => {
+  try {
+    const xcmTransfers: SubscanXCMTransferResult[] = []
+
+    for (const transfer of ongoingTransfers) {
+      const { crosschainMessageHash } = transfer
```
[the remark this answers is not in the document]
Victor-Salomon: Update everywhere 💁‍♂️

app/src/utils/snowbridge.ts
Victor-Salomon: Exactly. I just created a task on Linear as I found it good to have this in a dedicated PR and, once we have a clear vision of the parachain to AH tracking requirements.
Find it [here](https://linear.app/velocitylabs/issue/VEL-65/tracking-transfer-code-base-architecture-cleaning).

app/src/utils/snowbridge.ts
Victor-Salomon: Agreed with this as well. 
This should be covered by this [task](https://linear.app/velocitylabs/issue/VEL-65/tracking-transfer-code-base-architecture-cleaning).

app/src/utils/snowbridge.ts
Victor-Salomon: I will improve this for sure

app/src/hooks/useSnowbridgeTransferTracker.tsx
```diff
@@ -74,13 +93,11 @@ const useSnowbridgeTransferTracker = () => {
...
-        'extrinsic_hash' in transfer.submitted
-          ? transfer.submitted.extrinsic_hash === ongoing.id
-          : transfer.id === ongoing.id,
-      )
+
+      const foundTransfer = findOngoingTransfer(transfers, ongoing)
+
       if (foundTransfer) {
-        const msg = getStatusMessage(foundTransfer)
+        const msg = getTransferStatus(foundTransfer)
```
NunoAlexandre, line 100: Should the variable be renamed as well?

```suggestion
        const status = getTransferStatus(foundTransfer)
```

app/src/models/subscan.ts
```diff
@@ -0,0 +1,150 @@
+import { TransferStatus } from '@snowbridge/api/dist/history'
+
+export type SubscanEvent = {
```
NunoAlexandre, line 3: Is there a link to documentation where we can learn what these fields mean? for example, what is 'phase'?

app/src/models/subscan.ts
```diff
@@ -0,0 +1,150 @@
...
+
+export type SubscanParams = {
+  type: string
+  type_name: string
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  value: any
+  name: string
+}
+
+export type ParachainToETHTransferResult = {
```
NunoAlexandre, line 37: `nit` but it would be nice to rename `ETH` to `Eth` in these names  and abbreviate "Transfer" to `Tx` to make it easier to read.

```suggestion
export type ParachainToEthTxResult = {
```

Another suggestion, that I leave up to you., is to namespace these types so that the name doesn't have to be so extensive.

For example:

``` typescript 

namespace TxResult {
  type ParaToEth {
    ...
  }

  type SubscanXcm {
    ...
   }
}
```

And then we can reference those types a `TxResult.ParaToEth`, `TxResult.SubscanXcm`, etc.

app/src/models/subscan.ts
```diff
@@ -0,0 +1,150 @@
...
+    blockNumber: number
+    blockHash: string
+    transactionHash: string
+    transactionIndex: number
+    nonce: number
+    success: boolean
+  }
+}
+
+export type SubscanXCMTransferRawResponse = {
```
NunoAlexandre, line 86: We could follow the same idea of namespacing these "RawResponse" types.

app/src/utils/snowbridge.ts
```diff
@@ -83,9 +86,15 @@ export async function getTransferHistory(
     },
   }
 
-  const transfers: (ToEthereumTransferResult | ToPolkadotTransferResult)[] = []
+  const transfers: (
+    | ToEthereumTransferResult
+    | ToPolkadotTransferResult
+    | SubscanXCMTransferResult
```
NunoAlexandre, line 92: Linking my idea of the namespaces for these `TxResult` types, we could have a `type TxResultType = TxResult.ToEthereum | TxResult.ToPolkadot | TxResult.SubscanXcm`. Then here we could just reference that union type.

app/src/utils/snowbridge.ts
```diff
@@ -83,9 +86,15 @@ export async function getTransferHistory(
...
+  )[] = []
+
+  if (ongoingTransfers.toEthereum.fromAssetHub.length) {
```
NunoAlexandre, line 95: Does it matter if `length === 0`?

app/src/utils/snowbridge.ts
```diff
@@ -97,8 +106,24 @@ export async function getTransferHistory(
...
-    console.log('To Ethereum transfers:', toEthereum.length)
+    console.log('From AH To Ethereum transfers:', toEthereum.length)
     transfers.push(...toEthereum)
+
+    // const toEthereum = await trackXcmTransfer(
+    //   relaychainScan,
+    //   ongoingTransfers.toEthereum.fromAssetHub,
+    // )
+    // console.log('From AH To Ethereum transfer:', toEthereum.length)
+    // transfers.push(...toEthereum)
```
NunoAlexandre, line 117: Delete?

app/src/models/subscan.ts
```diff
@@ -0,0 +1,150 @@
...
+  assets: any
+}
+
+type SubscanXCMMetadata = {
+  sendAt?: number
+  txHash?: string
+  messageId?: string
+}
+
+export type SubscanXCMTransferResult = {
```
NunoAlexandre, line 130: btw do these types need to be prefixed with `Subscan`? As in, we use Subscan to track those txs but they are Xcm transfers, not Subscan-XCM transfers. Does that make sense? So if possible and logical we should not entangle the two things (the nature and protocol of the transfer with the tool used to track it)

NunoAlexandre (commented):
What a beast 🫨  Amazing work, Victor. I left a 💩 ton of comments but I want to emphasize my kudus for keeping your head straight through this whole process 🤟

app/src/app/api/history/route.ts
```diff
@@ -22,13 +22,20 @@ const getCachedTransferHistory = unstable_cache(
...
 
       ongoingTransfers.map(transfer => {
         switch (transfer.direction) {
           case Direction.ToEthereum: {
-            transfers.toEthereum.push(transfer)
+            if (transfer.sourceChain.chainId === env.config.ASSET_HUB_PARAID) {
+              transfers.toEthereum.fromAssetHub.push(transfer)
+            } else {
+              transfers.toEthereum.fromParachain.push(transfer)
+            }
```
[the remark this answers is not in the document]
Victor-Salomon: We will resolve/clean this code as soon as we can address [this task](https://linear.app/velocitylabs/issue/VEL-65/tracking-transfer-code-base-architecture-cleaning). (I'm adding this to the linear description).

The end goal is to track every transfer by messageHash (from Polkadot) or transactionHash (from Eth) using the same `trackXcmTransfer` helper 

In the meantime, we differentiate each tracking method until we're sure about both Subscan and AT API outcomes from Parachains (other than AH) to AH or directly to ETH transfers. Since we can't test it yet, I would advise to maintain it this way.

app/src/hooks/useAssetTransferApi.tsx
```diff
@@ -57,60 +57,104 @@ const useAssetTransferApi = () => {
...
-          token: token.name,
-          amount: amount.toString(),
-          destinationChain: destinationChain.name,
-          usdValue: tokenUSDValue,
-          usdFees: fees.inDollars,
-          recipient: recipient,
-          date: date.toISOString(),
+        .signAndSend(account.address, { signer: account.signer as any }, async result => {
+          // verify transaction hash & transfer isn't completed
+          if (!result.txHash) throw new Error('Transfer failed')
```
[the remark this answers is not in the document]
Victor-Salomon: I will update by: 'Transfer error: Failed to generate the transaction hash'

We could try to dig into the error/failed outcome. 
If I'm right, a missing _txHash_ should not happen unless we input some wrong parameters. Afair, a tx hash is valid for a few minutes only (5 mins if I recall correctly) to be signed and submitted.

app/src/hooks/useAssetTransferApi.tsx
```diff
@@ -57,60 +57,104 @@ const useAssetTransferApi = () => {
...
+          if (transferComplete) return
+
+          const isIncluded = result.status.isInBlock
+          // const isFinalized = result.status.isFinalized
```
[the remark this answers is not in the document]
Victor-Salomon: This is a small concern I have: should we wait for the transfer to be included in the block or wait for the block to be finalized? 

Once the user has signed the transfer, we wait for the txHash to be included in the block. The safest would be to wait for block finalization but I always used isInBlock. Block finalization will delay a lot more the flow which I do not like.

app/src/hooks/useAssetTransferApi.tsx
```diff
@@ -57,60 +57,104 @@ const useAssetTransferApi = () => {
...
+                messageHash = (data.messageHash as any).toString()
+              }
+              if (method === 'Sent' && section === 'polkadotXcm' && 'messageId' in data) {
+                // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                messageId = (data.messageId as any).toString()
+              }
+              if (method === 'ExtrinsicSuccess' && section === 'system') {
+                extrinsicSuccess = true
+              }
+            })
```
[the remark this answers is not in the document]
Victor-Salomon: > This will re-assign messageHash, messageId, and extrinsicSuccess every time the conditions are met. Do we want that? Or do we want to find the first even that satisfies those conditions? If the latter, we should use find instead of forEach

All three combinations of method/section are expected to be unique ([transfer example here](https://assethub-rococo.subscan.io/event?extrinsic=6231232-2)). I'm looking for the three events to consider the xcm transfer successful. If one of these events is missing, the transfer should not succeed. So far, from a perf pov, I guess one single loop with the forEach method is better than three loops with .find() one.  

> With that said, I'd suggest that we create a separate function to find the event we are looking for and returns a type containing messageHash, messageId, and extrinsicSuccess. That will drop 20+ lines of code from this context which will make this much more readable and easier to follow 👍

Agreed 💥

app/src/hooks/useAssetTransferApi.tsx
```diff
@@ -57,60 +57,104 @@ const useAssetTransferApi = () => {
...
+
+            if (!extrinsicSuccess)
+              throw new Error('Transfer failed. Returned extrinsicSuccess: false')
```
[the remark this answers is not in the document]
Victor-Salomon: As said before

Internal errors are not displayed to the user.
1- We do capture any of these errors in Sentry
2- The user will only have the generic error from the handleSendError() helper, whichaddNotification: 'Failed to submit the transfer'

_I'm not sure the user should receive a more advanced 'internal' error after successfully completing the form and passing the validation._ Also from a design pov, internal error not thrown by us could be too long?

app/src/utils/snowbridge.ts
```diff
@@ -154,34 +154,47 @@ export interface AccountInfo {
 export function getTransferStatus(
   transferResult: ToEthereumTransferResult | ToPolkadotTransferResult | SubscanXCMTransferResult,
 ) {
-  if ('info' in transferResult && transferResult.info.destinationParachain == undefined)
+  /** Retrieves the status of a transfer to ETH from a Snowbridge result */
```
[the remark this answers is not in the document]
Victor-Salomon: Updated ✅

app/src/hooks/useAssetTransferApi.tsx
```diff
@@ -57,60 +57,104 @@ const useAssetTransferApi = () => {
...
-          amount: amount.toString(),
-          destinationChain: destinationChain.name,
-          usdValue: tokenUSDValue,
-          usdFees: fees.inDollars,
-          recipient: recipient,
-          date: date.toISOString(),
+        .signAndSend(account.address, { signer: account.signer as any }, async result => {
+          // verify transaction hash & transfer isn't completed
+          if (!result.txHash) throw new Error('Transfer failed')
+          if (transferComplete) return
```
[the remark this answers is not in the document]
Victor-Salomon: See the response to the comment below ⬇️

app/src/hooks/useAssetTransferApi.tsx
```diff
@@ -57,60 +57,104 @@ const useAssetTransferApi = () => {
...
+                amount: amount.toString(),
+                destinationChain: destinationChain.name,
+                usdValue: tokenUSDValue,
+                usdFees: fees.inDollars,
+                recipient: recipient,
+                date: date.toISOString(),
+              })
+            }
+            // Mark the transfer as complete and return
+            transferComplete = true
```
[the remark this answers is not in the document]
Victor-Salomon: I initially implemented all the callback mechanisms with a _js new Promise()._ 
Unfortunately, the callback keeps looping depending on several events: For example, when the hash is in a finalized block, it will iterate again even tough I marked the promise as resolved. 

At least with the transferComplete, if it loops again, it exits the callback properly as you can see in the screenshot.

Rococo is super painful to test correctly. I can look again once we move to Paseo?

image

app/src/models/subscan.ts
Victor-Salomon: Unfortunately, there is no proper documentation on this. 
I mostly get them from logging the response. Subscan API does not provide any types. 

Phase is part of the events returned in an extrinsic result.

app/src/models/subscan.ts
Victor-Salomon: This is nice.

_I'm adding this lead to the cleaning part in [this task](https://linear.app/velocitylabs/issue/VEL-65/tracking-transfer-code-base-architecture-cleaning)._

app/src/models/subscan.ts
Victor-Salomon: Same this will be handled in [this task](https://linear.app/velocitylabs/issue/VEL-65/tracking-transfer-code-base-architecture-cleaning).

app/src/utils/snowbridge.ts
Victor-Salomon: Same this will be handled in [this task](https://linear.app/velocitylabs/issue/VEL-65/tracking-transfer-code-base-architecture-cleaning).

=> We should get rid of the two Snowbridge responses to keep only one : SubscanXCMTransferResult

app/src/utils/snowbridge.ts
Victor-Salomon: Yes, it does:
 If .length === 0, it will be considered falsy and will not execute, while if I do not check the length, the empty [] array will exist.

app/src/utils/snowbridge.ts
Victor-Salomon: We should switch to this once Subscan confirms the stable version. Let's keep it for now.

NunoAlexandre (approved):
👏 🥐 ✨

noahjoeris (approved):

PR #122, opened 2024-09-06
Show beta warning for transfers going through AT API

NunoAlexandre:
### Changes

- Drop `useDryRunValidation` - The PJS validation is not useful enough to justify a two signature flow
- [off-Improve naming of the from/to swap feature
- Make ActionBanner more flexible by making the button optional
- ⭐  Show banner with warning for transfers that will be run through AT API


app/public/wip.png
Victor-Salomon, line 1: 🤩

app/src/components/Transfer.tsx
```diff
@@ -1,14 +1,13 @@
...
 import useErc20Allowance from '@/hooks/useErc20Allowance'
 import useSnowbridgeContext from '@/hooks/useSnowbridgeContext'
 import useTransferForm from '@/hooks/useTransferForm'
 import { resolveDirection } from '@/services/transfer'
-import { getRecipientAddress } from '@/utils/address'
 import {
   getAllowedDestinationChains,
   getAllowedSourceChains,
   getAllowedTokens,
+  getRoute,
```
Victor-Salomon, line 10: Let's be careful with file namings that use the word: `route`
For example: We have a `routes.ts `in utils. 

**`route`** is a word "owned" by Next.js as part of its architecture. 

There is no proper issue for now. 
This is more like a global warning to keep in mind :)

Victor-Salomon (approved):
Looks good to me. 
Warning text is simple and smart 👌

app/src/components/Transfer.tsx
NunoAlexandre: I agree!

PR #124, opened 2024-09-10
feat: Eth to wETH swap

NunoAlexandre:
When the user wishes to transfer a given amount of wETH but they either don't have balance or it's below the amount they wish to send, AND they have enough ETH balance to make that transfer happen, we now offer an ActionBanner to let the user swap ETH for wETH.

### TO DO

- [ ] Fix an issue where the validation doesn't update once the swap is finalised even tho we verify that the wETH is now sufficient


app/src/components/Transfer.tsx
```diff
@@ -69,6 +71,32 @@ const Transfer: FC = () => {
...
+    swapEthtoWEth,
+    swapping: isSwappingEthForWEth,
+  } = useEthForWEthSwap({
+    context: snowbridgeContext,
+    network: sourceChain?.network,
+    tokenAmount,
+    owner: sourceWallet?.sender?.address,
+  })
+
+  const shouldDisplayEthToWEthSwap =
```
Victor-Salomon, line 85: This might return something different than a boolean: 
image

app/src/components/Transfer.tsx
```diff
@@ -69,6 +72,38 @@ const Transfer: FC = () => {
...
+    tokenAmount?.token?.symbol === 'wETH' &&
+    !!tokenAmount?.amount &&
+    !!balanceData &&
+    !!ethBalance &&
+    // The user wants to send more than the balance available
+    tokenAmount.amount > Number(balanceData.formatted) &&
+    // but they have enough ETH to make it possible
+    tokenAmount.amount - Number(balanceData.formatted) < ethBalance &&
+    // We don't want two ActionBanners showing up at once
+    !requiresErc20SpendApproval
```
Victor-Salomon, line 101: I noticed this while testing yesterday, but wasn't sure if it was intentional! 
Nice to avoid the double banners 💥!

app/src/hooks/useErc20Balance.tsx
```diff
@@ -26,10 +26,13 @@ const useErc20Balance = ({ network, token, address, context }: UseBalanceParams)
     try {
       setLoading(true)
       let fetchedBalance: Erc20Balance | undefined
+      console.log('Getting wETH balance')
```
Victor-Salomon, line 29: Could we remove the logs here?

app/src/hooks/useEthForWEthSwap.tsx
```diff
@@ -0,0 +1,110 @@
...
+const useEthForWEthSwap = ({ network, tokenAmount, owner, context }: Params) => {
+  const { addNotification } = useNotification()
+  const { data: tokenBalance } = useErc20Balance({
+    network,
+    token: tokenAmount?.token ?? undefined,
+    address: owner,
+    context,
+  })
+  const [ethBalance, setEthBalance] = useState<number | undefined>()
+  const [swapping, setSwapping] = useState<boolean>(false)
```
Victor-Salomon, line 32: convention for loading state is to prefix a 'is' => isSwapping, setIsSwapping. 
But it does not really matter ;)

Victor-Salomon (approved):
Nice PR @NunoAlexandre 🦾.

app/src/components/Transfer.tsx
NunoAlexandre: Exactly! Maybe we will need a more scalable approach in the future but let's see how this rolls for now

PR #125, opened 2024-09-12
fix: Preserve Transfer form state on menu tab switch

NunoAlexandre:
Fixes https://linear.app/velocitylabs/issue/VEL-60/preserve-transfer-form-state-when-switching-to-completed-transfers

### Changes 
Basically, instead of rendering or removing the form and ongoing txs components based on the selected tab, we show or hide it, respectively.

### DIFF
--- a/app/src/components/HomeComponentSelect.tsx
+++ b/app/src/components/HomeComponentSelect.tsx
@@ -31,7 +31,7 @@ export const HomeComponentSelect = () => {
       <div
         className={cn(
           'z-15 relative max-w-[90vw]',
-          newTransferInit !== TransferTab.New ? 'hidden' : '',
+          newTransferInit !== TransferTab.New && 'hidden',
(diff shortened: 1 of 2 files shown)


app/src/components/HomeComponentSelect.tsx
```diff
@@ -26,18 +27,23 @@ export const HomeComponentSelect = () => {
...
-          />
-        </div>
-      ) : (
-        hasCompletedTransfers &&
-        completedTransfers && <TransferHistory transactions={completedTransfers} />
+
+      <div
+        className={cn(
+          'z-15 relative max-w-[90vw]',
+          newTransferInit !== TransferTab.New ? 'hidden' : '',
```
noahjoeris, line 34:
```suggestion
          newTransferInit !== TransferTab.New && 'hidden',
```

PR #126, opened 2024-09-16
Fix completed txs showing on New tab

NunoAlexandre:
Fixes https://linear.app/velocitylabs/issue/VEL-62/fix-completed-transfers-are-displayed-under-the-transfer-component


Victor-Salomon (approved):
All good here. 
Tested locally 👌

PR #132, opened 2024-09-26
feat: Support wETH transfers from/to Bifrost

NunoAlexandre:


app/src/utils/env.ts
```diff
@@ -3,7 +3,7 @@ export const isPreview: boolean = process.env.VERCEL_ENV === 'preview'
 export const projectId: string | undefined = process.env.NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID
 export const vercelDomain: string | undefined = process.env.NEXT_PUBLIC_VERCEL_URL
 
-export const shouldUseTestnet = isDevelopment || isPreview
+export const shouldUseTestnet = false // isDevelopment || isPreview
```
NunoAlexandre, line 6: Undo before merging - and we should find a better way to do this

noahjoeris:
I added the dry run and everything should work once AT API guys fix the bifrost issue. Let's see.

app/src/hooks/useAssetTransferApi.tsx
```diff
@@ -32,41 +33,30 @@ const useAssetTransferApi = () => {
...
-        // asset id
-        [token.multilocation],
-        // the amount (pairs with the asset ids above)
-        [amount.toString()],
-        {
-          format: 'submittable',
-          xcmVersion: safeXcmVersion,
-        },
-      )
+      const dryRunResult = await validate(atApi, params, setStatus)
```
Victor-Salomon, line 43: I find the namings here confusing. 

- `validate()` could be named `validateAssetTransferTx(...)`? 
-  `dryRunResult` could be kept as `txResult.` However, to keep the dry run explicit, what about having a new boolean constant (for example, `allowDryRun: true`) to pass as a param to the validate()/validateAssetTransferTx(allowDryRun)? 

Just an idea/reflexion.

app/src/hooks/useAssetTransferApi.tsx
```diff
@@ -32,41 +33,30 @@ const useAssetTransferApi = () => {
...
-        .tx(txResult.tx)
+        .tx(dryRunResult.tx)
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         .signAndSend(account.address, { signer: account.signer as any }, async result => {
           // verify transaction hash & transfer isn't completed
           if (!result.txHash)
             throw new Error('Transfer error: Failed to generate the transaction hash')
           if (isComplete) return
 
+          console.log('Transfer result:', result)
```
Victor-Salomon, line 59: Before merging, let's remove the logs

app/src/hooks/useFees.tsx
```diff
@@ -37,24 +47,56 @@ const useFees = (
       let tokenUSDValue: number = 0
       switch (direction) {
         case Direction.ToEthereum: {
-          const dotUSDValue = await getFeesTokenUSDValue(Network.Polkadot)
-          tokenUSDValue = dotUSDValue?.[Network.Polkadot?.toLowerCase()]?.usd ?? 0
+          if (!snowbridgeContext) throw new Error('Snowbridge context undefined')
```
Victor-Salomon, line 50: Why not move the if (!snowbridgeContext) before the switch and avoid duplicate checks?

app/src/utils/papi.ts
```diff
@@ -87,16 +92,44 @@ export const getNativeBalance = async (
   return await apiAssetHub?.query.System.Account.getValue(address)
 }
```
Victor-Salomon, line 94: Should we move the `ethMultilocationSchema` in this file to the `models/schemas.ts` file?

Victor-Salomon (commented):
A very nice job to you guys! 
Not far from being ready to be released. 
I've left a couple of comments. 

Also, the build has failed due to a type/any issue in the useSnowbridgeApi() hook. 

Let's wait for the AT API team to release the patch and test all of this 💥

app/src/hooks/useFees.tsx
NunoAlexandre: I moved it here because we were checking it even if the `Direction` would turn out to be `WithinPolkadot`. That would mean that we would fail to load tx fees for an XCM transfer because of the snowbridge context which is terrible

NunoAlexandre (commented):
Almost there 🏄  Really cool that the latest AT API fixed our blockage!!

app/src/config/registry.ts
```diff
@@ -313,6 +313,10 @@ export namespace Testnet {
 
 export type TransferSDK = 'AssetTransferApi' | 'SnowbridgeApi'
 
+export type AssetIds = {
+  Token2: number
+}
+
```
NunoAlexandre, line 319: @noahjoeris the idea is good but in some instances the asset id will be just a string or just a number and then this won't work. Besides, `Token2` is Bifrost-depandant, it won't be usable for other chains.

NunoAlexandre: I dropped the whole `localAssetId` so this is all gone 👍

PR #134, opened 2024-10-01
feat: Add loading balance indicator

noahjoeris:
This PR adds a simple loading placeholder if the balance is fetching


NunoAlexandre (approved):
✨

Issue #109, opened 2024-08-23
Text still looks really big


brandonhaslegs:
Did you make the biggest headline text size use `vw` amounts? This headline looks way too big.

<img width="1171" alt="image" src="https://github.com/user-attachments/assets/863a5bb0-aba5-43b8-903d-db433113ba7a">

PR #135, opened 2024-10-02
Fix landing page text styling

rudolfs:
Fixes https://github.com/velocitylabs-org/turtle/issues/109. As well as the word break issue @brandonhaslegs reported.

https://github.com/user-attachments/assets/f857e53a-5325-4686-802c-38b98c8362d7

### DIFF
--- a/site/src/components/card.tsx
+++ b/site/src/components/card.tsx
@@ -36,9 +36,11 @@ export const FeatureCard = ({
         width={126}
         className="rounded-[26px] bg-secondary"
       />
-      <h3 className="max-w-xl break-words text-2xl font-medium sm:text-4xl xl:text-h-sub xl:leading-12 ">
-        {feature.title}
-      </h3>
+      <h3
+        lang="en"
+        className="max-w-xl hyphens-auto text-2xl font-medium sm:text-4xl xl:text-h-sub xl:leading-12"
+        dangerouslySetInnerHTML={{ __html: feature.title }}
+      ></h3>
       <p className="text-sm font-medium tracking-[0.03rem] sm:w-5/6">{feature.description}</p>
       {feature.tags && (
         <div className="flex space-x-1 sm:space-x-2">
--- a/site/src/components/features.tsx
+++ b/site/src/components/features.tsx
@@ -13,7 +13,7 @@ export const Features = () => {
       icon: turtle,
     },
     {
-      title: 'Transparency all the way',
+      title: 'Trans&shy;parency all the way',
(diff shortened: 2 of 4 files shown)


site/src/components/features.tsx
```diff
@@ -13,7 +13,7 @@ export const Features = () => {
       icon: turtle,
     },
     {
-      title: 'Transparency all the way',
+      title: 'Trans&shy;parency all the way',
```
rudolfs, line 16: Turns out capitalized words aren't hyphenated by default by browsers in English, hence this workaround. Open for other suggestions on how to fix this in a nicer way.

noahjoeris (approved):

site/src/components/card.tsx
```diff
@@ -36,9 +36,11 @@ export const FeatureCard = ({
         width={126}
         className="rounded-[26px] bg-secondary"
       />
-      <h3 className="max-w-xl break-words text-2xl font-medium sm:text-4xl xl:text-h-sub xl:leading-12 ">
-        {feature.title}
-      </h3>
+      <h3
+        lang="en"
+        className="max-w-xl hyphens-auto text-2xl font-medium sm:text-4xl xl:text-h-sub xl:leading-12"
+        dangerouslySetInnerHTML={{ __html: feature.title }}
```
NunoAlexandre, line 42: why do we need `dangerouslySetInnerHTML` here? is it because it's printing an injected value?

site/tailwind.config.ts
```diff
@@ -69,7 +69,8 @@ const config: Config = {
         sm: '0.03em',
       },
       screens: {
-        '3xl': '2100px',
+        '3xl': '1730px',
+        '4xl': '2100px',
```
NunoAlexandre, line 73: 🫡

site/src/components/card.tsx
rudolfs: Yes, that's the reason. Otherwise it will render the `&shy;` literally. It should be safe since we're not passing user input, but kinda ugly still.

site/tailwind.config.ts
rudolfs: `3xl` wasn't used anywhere as far as I could tell, so should be fine to rename.

site/src/components/card.tsx
NunoAlexandre: I see. good hack tbh, I'm fine with this approach 👍

NunoAlexandre (approved):

PR #136, opened 2024-10-02
dx: Set a nodejs version for people using asdf

rudolfs:
I'm using a runtime manager, so this would make my life easier.


noahjoeris (approved):
Sure, I don't mind adding it. 👍

NunoAlexandre (approved):
👌

PR #139, opened 2024-10-04
Improvement: tracking type refactoring

Victor-Salomon:
This PR contains the **second** part of the transfer tracking cleaning & improvements: Types

Again, no big logic change.

- API route now supports WithinPolkadot transfer direction. 
- TxTrackingResult type provides more clarity within the code.


app/src/models/subscan.ts
```diff
@@ -148,3 +65,89 @@ export type SubscanXCMTransferResult = {
   metadata: SubscanXCMMetadata
   status: TransferStatus
 }
+
+// LEGACY CODE BELOW
+// TO KEEP UNTIL FURTHER NOTICE
```
NunoAlexandre, line 70: Why do we want to keep this commented code?

app/src/models/transfer.ts
```diff
@@ -88,3 +87,5 @@ export enum TransferTab {
   Completed = 'Completed',
 }
 export type TransferTabOptions = TransferTab
+
+export type TxTrackingResult = FromEthTrackingRes | FromParachainTrackingRes
```
NunoAlexandre, line 91: Can we be consistent with the usage of `Result` vs `Res`? 🙏

app/src/models/subscan.ts
Victor-Salomon: I want to keep it until we test Subscan endpoint support properly Parachain to AH transfers.

app/src/models/transfer.ts
Victor-Salomon: Obviously :)

app/src/models/subscan.ts
noahjoeris: Usually I would say it is in VCS anyways, but with Squash Merge I am actually not sure 😄

Victor-Salomon: You're right 😁
Let me remove it.

noahjoeris (approved):
Great 🚀 Just the naming stuff Nuno mentioned above.

Issue #94, opened 2024-08-13
Add a #fff background to all avatars


brandonhaslegs:
<img width="343" alt="image" src="https://github.com/user-attachments/assets/bb8f0775-624f-4402-b204-d51b28cea144">

So that if you have a transparent PNG avatar, it doesn't take on the background color, like the Rococo Asset Hub avatar icon is doing in this screenshot.

PR #144, opened 2024-10-09
fix: Add a background color to all avatars

rudolfs:
Closes https://github.com/velocitylabs-org/turtle/issues/94

NB: I didn't visually check all the instances that I changed. We should think about visual regression tests for the app.

Screenshot 2024-10-09 at 10 02 18

### DIFF
--- a/app/src/components/ChainSelect.tsx
+++ b/app/src/components/ChainSelect.tsx
@@ -93,7 +93,7 @@ const ChainSelect = forwardRef<HTMLDivElement, ChainSelectProps>(
                     alt={value.name}
                     width={24}
                     height={24}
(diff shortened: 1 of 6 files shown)


noahjoeris (approved):
LGTM 👍

NunoAlexandre:
> NB: I didn't visually check all the instances that I changed. We should think about visual regression tests for the app.


@rudolfs any suggestions on how to approach that?

NunoAlexandre (approved):

rudolfs:
> > NB: I didn't visually check all the instances that I changed. We should think about visual regression tests for the app.
> 
> @rudolfs any suggestions on how to approach that?

Yeah, I have some ideas that we can discuss.

PR #145, opened 2024-10-09
improvement: explorer link update

Victor-Salomon:
This PR Support displaying tracking through the XCM transfers Subscan Relaychain tracking page: 
- add ongoingTransfer Store method to update the UniqueId
- optional param addition to StoredTransfer type
- implement explorer link in getExplorerLink when uniqueId is available.

### DIFF
--- a/app/src/store/ongoingTransfersStore.ts
+++ b/app/src/store/ongoingTransfersStore.ts
@@ -63,18 +63,13 @@ export const useOngoingTransfersStore = create<State>()(
       updateTransferUniqueId: (id: string, uniqueTrackingId: string) => {
         if (!id || !uniqueTrackingId) return
         set(state => {
-          // Check if the newOngoingTransfer already exists in the local storage
-          if (state.transfers.some(transfer => transfer.uniqueTrackingId === uniqueTrackingId))
-            return state
-
-          state.transfers.map(transfer => {
-            if (transfer.id == id) {
-              transfer.uniqueTrackingId = uniqueTrackingId
-            }
-          })
-
           return {
-            transfers: [...state.transfers],
+            transfers: state.transfers.map(transfer => {
+              if (transfer.id == id) {
+                transfer.uniqueTrackingId = uniqueTrackingId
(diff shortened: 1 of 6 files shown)


app/src/store/ongoingTransfersStore.ts
```diff
@@ -59,6 +60,25 @@ export const useOngoingTransfersStore = create<State>()(
         })
       },
 
+      updateTransferUniqueId: (id: string, uniqueTrackingId: string) => {
+        if (!id || !uniqueTrackingId) return
+        set(state => {
+          // Check if the newOngoingTransfer already exists in the local storage
+          if (state.transfers.some(transfer => transfer.uniqueTrackingId === uniqueTrackingId))
+            return state
```
NunoAlexandre, line 68: We should drop this. The reason for that is that now we first do O(n), going through all the transfers to check if we already have one with that `uniqueTrackingId`.

Then we go through all the transfers again to check `if (transfer.id == id)` and change the `uniqueTrackingId` of that tx if found.

So this can just be:

return {
  transfers: state.transfers.map(tx => if (transfer.id == id) {  transfer.uniqueTrackingId = uniqueTrackingId })
}

app/src/utils/transfer.ts
```diff
@@ -119,27 +119,41 @@ export function getExplorerLink(transfer: StoredTransfer): string | undefined {
...
       console.log(`Unsupported network: ${network}`)
   }
 }
 
+export const getSubdomainPath = (url: string) => {
+  const parsedUrl = new URL(url)
+  const hostname = parsedUrl.hostname
+  return hostname.split('.')[0]
+}
+
```
NunoAlexandre, line 156: Can we document this fn? :)

NunoAlexandre (commented):
Two comments, looking good already :)

app/src/utils/transfer.ts
Victor-Salomon: obviously! Let me add it :)

PR #146, opened 2024-10-09
fix: transfer data callback handler & subscan loop

Victor-Salomon:
This PR fix the following: 

- The transfer data handling within the AT API callback: switch transfer result status check from isInBlock to isFinalized. It should prevent registering a first bunch of data that changes once block data is finalized.

- Fixes Subscan, exiting loop too early when an ongoing transfer fails.

### DIFF
--- a/app/src/hooks/useAssetTransferApi.tsx
+++ b/app/src/hooks/useAssetTransferApi.tsx
@@ -68,7 +68,7 @@ const useAssetTransferApi = () => {
           if (!txHash) throw new Error('Transfer error: Failed to generate the transaction hash')
           if (isComplete) return
 
-          // Wait until block to be finalized before handling transfer data
+          // Wait until block is finalized before handling transfer data
           if (status.isFinalized) {
             let messageHash: string | undefined
             let messageId: string | undefined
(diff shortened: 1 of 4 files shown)


app/src/hooks/useAssetTransferApi.tsx
```diff
@@ -62,21 +62,20 @@ const useAssetTransferApi = () => {
...
-          if (!result.txHash)
-            throw new Error('Transfer error: Failed to generate the transaction hash')
+          if (!txHash) throw new Error('Transfer error: Failed to generate the transaction hash')
           if (isComplete) return
 
-          const isIncluded = result.status.isInBlock
-          // const isFinalized = result.status.isFinalized
-          if (isIncluded) {
-            // if (isIncluded || isFinalized) {
+          // Wait until block to be finalized before handling transfer data
```
NunoAlexandre, line 71:
```suggestion
          // Wait until block is finalized before handling transfer data
```

NunoAlexandre (approved):
If it works, it looks good to me 👍

noahjoeris (approved):

PR #147, opened 2024-10-11
refactor: config

noahjoeris:
- let's us specify the environment via env variables now
- adds the groundwork to support switching environments inside the app in the future

### DIFF
--- a/app/.env.local.example
+++ b/app/.env.local.example
@@ -35,3 +35,5 @@ NEXT_PUBLIC_POLKADOT_RELAY_CHAIN_API_URL='https://polkadot-rpc.dwellir.com'
 # Parachain rpc endpoints. Used for snowbridge context and it uses a public node. Change to private nodes if needed
 NEXT_PUBLIC_PARACHAIN_API_URLS='["https://polkadot-mythos-rpc.polkadot.io", "https://moonbeam-rpc.dwellir.com", "https://hydradx-rpc.dwellir.com", "https://bifrost-rpc.dwellir.com"]'
 
+# Inits the environment of the app. Change to 'testnet' if needed.
+NEXT_PUBLIC_ENVIRONMENT='mainnet'
--- a/app/src/app/api/history/route.ts
+++ b/app/src/app/api/history/route.ts
@@ -3,21 +3,20 @@ export const dynamic = 'force-dynamic' // Always run dynamically
 export const revalidate = 30 // Keep cache for 2 minutes
 export const maxDuration = 90 // Timout adter
 
-import { unstable_cache } from 'next/cache'
-import { NextResponse } from 'next/server'
 import { getEnvironment } from '@/context/snowbridge'
 import { ongoingTransfersSchema } from '@/models/api-schemas'
 import { OngoingTransferWithDirection, OngoingTransfers } from '@/models/transfer'
 import { Direction } from '@/services/transfer'
-import { Environment } from '@/store/environmentStore'
-import { shouldUseTestnet } from '@/utils/env'
+import { Environment, environmentFromStr } from '@/store/environmentStore'
 import { getErrorMessage, trackTransfers } from '@/utils/transferTracking'
+import { unstable_cache } from 'next/cache'
+import { NextResponse } from 'next/server'
 
 const CACHE_REVALIDATE_IN_SECONDS = 30
 
 const getCachedTransferHistory = unstable_cache(
-  (ongoingTransfers: OngoingTransferWithDirection[]) => {
-    const env = getEnvironment(shouldUseTestnet ? Environment.Testnet : Environment.Mainnet)
+  (env: Environment, ongoingTransfers: OngoingTransferWithDirection[]) => {
+    const snowbridgeEnv = getEnvironment(env)
 
     try {
       const transfers: OngoingTransfers = {
@@ -29,7 +28,7 @@ const getCachedTransferHistory = unstable_cache(
       ongoingTransfers.map(transfer => {
         switch (transfer.direction) {
           case Direction.ToEthereum: {
-            if (transfer.sourceChain.chainId === env.config.ASSET_HUB_PARAID) {
+            if (transfer.sourceChain.chainId === snowbridgeEnv.config.ASSET_HUB_PARAID) {
               transfers.toEthereum.push(transfer)
             } else {
               console.log('Direct Parachain => Eth tracking to be implemented/verified')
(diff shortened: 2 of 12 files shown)


app/.env.local.example
```diff
@@ -35,3 +35,5 @@ NEXT_PUBLIC_POLKADOT_RELAY_CHAIN_API_URL='https://polkadot-rpc.dwellir.com'
 # Parachain rpc endpoints. Used for snowbridge context and it uses a public node. Change to private nodes if needed
 NEXT_PUBLIC_PARACHAIN_API_URLS='["https://polkadot-mythos-rpc.polkadot.io", "https://moonbeam-rpc.dwellir.com", "https://hydradx-rpc.dwellir.com", "https://bifrost-rpc.dwellir.com"]'
 
+# Inits the environment of the app. Change to 'testnet' if needed.
+NEXT_PUBLIC_ENVIRONMENT='mainnet'
```
NunoAlexandre, line 39: Question: let's say we have another NEXT JS project on the same machine, these envs will clash. Why is it not common standard to prefix these vars with the name of the project? So `TURTLE_APP_ENV` for example

app/src/app/api/history/route.ts
```diff
@@ -3,21 +3,20 @@ export const dynamic = 'force-dynamic' // Always run dynamically
...
+import { unstable_cache } from 'next/cache'
+import { NextResponse } from 'next/server'
 
 const CACHE_REVALIDATE_IN_SECONDS = 30
 
 const getCachedTransferHistory = unstable_cache(
-  (ongoingTransfers: OngoingTransferWithDirection[]) => {
-    const env = getEnvironment(shouldUseTestnet ? Environment.Testnet : Environment.Mainnet)
+  (env: Environment, ongoingTransfers: OngoingTransferWithDirection[]) => {
+    const snowbridgeEnv = getEnvironment(env)
```
NunoAlexandre, line 19: 👌  way to go

app/src/app/api/history/route.ts
```diff
@@ -29,7 +28,7 @@ const getCachedTransferHistory = unstable_cache(
       ongoingTransfers.map(transfer => {
         switch (transfer.direction) {
           case Direction.ToEthereum: {
-            if (transfer.sourceChain.chainId === env.config.ASSET_HUB_PARAID) {
+            if (transfer.sourceChain.chainId === snowbridgeEnv.config.ASSET_HUB_PARAID) {
```
NunoAlexandre, line 31: 👌

app/src/config/index.tsx
```diff
@@ -21,7 +21,8 @@ const chainsMainnet = [mainnet] as const
 
 // Create wagmiConfig
 export const config = defaultWagmiConfig({
-  chains: shouldUseTestnet ? chainsTestnet : chainsMainnet,
+  chains:
+    process.env.NEXT_PUBLIC_ENVIRONMENT === Environment.Testnet ? chainsTestnet : chainsMainnet, // TODO: Figure out how to access envirnoment state here once needed.
```
NunoAlexandre, line 25:
```suggestion
    process.env.NEXT_PUBLIC_ENVIRONMENT === Environment.Testnet ? chainsTestnet : chainsMainnet, // TODO: Figure out how to access environment state here once needed.
```

and what do you mean by "environment state"?

app/src/hooks/useEnvironment.tsx
```diff
@@ -1,9 +1,8 @@
-import { Environment } from '@/store/environmentStore'
-import { shouldUseTestnet } from '@/utils/env'
+import { useEnvironmentStore } from '@/store/environmentStore'
 
 const useEnvironment = () => {
-  const environment = shouldUseTestnet ? Environment.Testnet : Environment.Mainnet
-  return environment
+  const current = useEnvironmentStore(state => state.current)
+  return current
```
NunoAlexandre, line 5:
```suggestion
  return useEnvironmentStore(state => state.current)
```

if this works, it's cleaner imo

app/src/hooks/useOngoingTransfersTracker.tsx
```diff
@@ -57,7 +56,7 @@ const useOngoingTransfersTracker = () => {
 
     try {
       setLoading(true)
-      const response = await fetch('/api/history', {
+      const response = await fetch(`/api/history?env=${env}`, {
```
NunoAlexandre, line 59: good change 👍

app/src/store/environmentStore.ts
```diff
@@ -26,13 +28,13 @@ export const useEnvironmentStore = create<State>(set => ({
...
-export function environmentFromStr(str: string): Environment | undefined {
+export function environmentFromStr(str: string): Environment {
   switch (str.toLowerCase()) {
     case 'mainnet':
       return Environment.Mainnet
     case 'testnet':
       return Environment.Testnet
     default:
-      return
+      return Environment.Mainnet // default to mainnet
```
NunoAlexandre, line 38:
```suggestion
      return Environment.Mainnet
```

The code already states that clearly 👍

NunoAlexandre (commented):
Minor feedback, overall a really sweet refactor 👌

app/src/context/index.tsx
```diff
@@ -19,7 +17,6 @@ export const wallet = createWeb3Modal({
   projectId,
   enableAnalytics: true,
   enableOnramp: true,
-  defaultChain: shouldUseTestnet ? sepolia : mainnet,
```
Victor-Salomon, line 22: Was it this param that created the confusion in the Vercel development/preview?

Victor-Salomon (approved):
Nice PR @noahjoeris! 
I really like the update of the route API 😍
I left one question which does not require a code change.

app/.env.local.example
noahjoeris: As long as we host it on Vercel it doesn't matter. Env variables are separated per project. So we don't need that prefix.

app/src/config/index.tsx
noahjoeris: That's an obstacle I couldn't find a solution for right now. Maybe @Victor-Salomon has an idea. 
My goal was to store the environment in Zustand state management and use it from there. Here I just load the env variable and it doesnt adapt yet to the state. So no full in-app environment change possible yet.
The reason is that hooks can only be used inside components and not in such a config file.

app/src/context/index.tsx
noahjoeris: What confusion do you mean? 
The defaultChain just defaults to a particular chain. So Sepolia or Ethereum.

noahjoeris:
Thanks for the review guys 🐰

app/src/config/index.tsx
Victor-Salomon: Right now, this config is built on the server side. 

Don't you think we could move it directly to =>` src/context/index.ts` and handle the switch inside this? Since it's a client component, maybe this is doable. 
This needs anyway to be investigated further.

noahjoeris: Okay let's try that once we need it.

PR #148, opened 2024-10-11
Fix/tracking from snowbridge api

Victor-Salomon:
This PR revert some changes to support AH to Eth transfer using Snowbridge API. 

The `trackFromAhToEthTx()` is the same code as we used before but in a dedicated helper. 
It contains more documentation about helpers as well.

### DIFF
--- a/app/src/models/snowbridge.ts
+++ b/app/src/models/snowbridge.ts
@@ -6,3 +6,4 @@ export type SnowbridgeStatus = {
 }
 
 export type FromEthTrackingResult = history.ToPolkadotTransferResult
+export type FromAhToEthTrackingResult = history.ToEthereumTransferResult
--- a/app/src/utils/snowbridge.ts
+++ b/app/src/utils/snowbridge.ts
@@ -1,7 +1,7 @@
 import { AlchemyProvider } from 'ethers'
 import { environment, subscan, history } from '@snowbridge/api'
-import { IGateway__factory } from '@snowbridge/contract-types'
-import { FromEthTrackingResult } from '@/models/snowbridge'
+import { BeefyClient__factory, IGateway__factory } from '@snowbridge/contract-types'
+import { FromAhToEthTrackingResult, FromEthTrackingResult } from '@/models/snowbridge'
 
 export const SKIP_LIGHT_CLIENT_UPDATES = true
 export const HISTORY_IN_SECONDS = 60 * 60 * 24 * 2 // 2 days
@@ -48,9 +48,7 @@ export async function trackFromEthTx(
     utcNowTimestamp - historyInSeconds,
   )
 
-  if (ethNowBlock === null) {
-    throw Error('Could not fetch latest Ethereum block.')
-  }
+  if (ethNowBlock === null) throw Error('Could not fetch latest Ethereum block.')
 
   const searchRange = {
     assetHub: {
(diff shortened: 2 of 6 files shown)


app/src/models/snowbridge.ts
```diff
@@ -6,3 +6,4 @@ export type SnowbridgeStatus = {
 }
 
 export type FromEthTrackingResult = history.ToPolkadotTransferResult
+export type FromAhToEthTrackingResult = history.ToEthereumTransferResult
```
NunoAlexandre, line 9: It would be nice to namespace these and maybe even simplify the naming so as to have `Result.AhToEth`. Not a blocker at all, just a reminder of this same comment I had left some weeks back

app/src/utils/snowbridge.ts
```diff
@@ -1,10 +1,10 @@
 import { AlchemyProvider } from 'ethers'
 import { environment, subscan, history } from '@snowbridge/api'
-import { IGateway__factory } from '@snowbridge/contract-types'
-import { FromEthTrackingResult } from '@/models/snowbridge'
+import { BeefyClient__factory, IGateway__factory } from '@snowbridge/contract-types'
+import { FromAhToEthTrackingResult, FromEthTrackingResult } from '@/models/snowbridge'
 
 export const SKIP_LIGHT_CLIENT_UPDATES = true
-export const HISTORY_IN_SECONDS = 60 * 60 * 24 * 2 // 2 days
+export const HISTORY_IN_SECONDS = 60 * 60 * 24 * 5 // 5 days
```
NunoAlexandre, line 7: This means we will check for tracking updates in the last 5 days, right?

app/src/utils/snowbridge.ts
```diff
@@ -79,3 +77,74 @@ export async function trackFromEthTx(
...
   return transfers
 }
+
+export async function trackFromAhToEthTx(
+  env: environment.SnowbridgeEnvironment,
+  skipLightClientUpdates = SKIP_LIGHT_CLIENT_UPDATES,
+  historyInSeconds = HISTORY_IN_SECONDS,
+): Promise<FromAhToEthTrackingResult[]> {
+  if (!env.config.SUBSCAN_API) {
+    console.warn(`No subscan api urls configured for ${env.name}`)
```
NunoAlexandre, line 87: Should we return an error? I'd say this is important to track with Sentry

app/src/utils/snowbridge.ts
```diff
@@ -79,3 +77,74 @@ export async function trackFromEthTx(
...
+
+  const subscanKey = process.env.NEXT_PUBLIC_SUBSCAN_KEY
+  if (!subscanKey) throw Error('Missing Subscan Key')
+
+  const ethereumProvider = new AlchemyProvider(env.ethChainId, alchemyKey)
+
+  const assetHubScan = subscan.createApi(env.config.SUBSCAN_API.ASSET_HUB_URL, subscanKey)
+  const bridgeHubScan = subscan.createApi(env.config.SUBSCAN_API.BRIDGE_HUB_URL, subscanKey)
+  const relaychainScan = subscan.createApi(env.config.SUBSCAN_API.RELAY_CHAIN_URL, subscanKey)
+  const assetHubParaId = env.config.ASSET_HUB_PARAID
```
NunoAlexandre, line 101: Shouldn't we check that this is set? And I'd say we should get this from the registry, not from the environment

app/src/utils/transferTracking.ts
```diff
@@ -35,31 +34,80 @@ export const trackTransfers = async (
...
+ * The transfer direction is from AH to Eth (from Snowbridge API)
+ *
+ * If a result includes 'destEventIndex' in the transferResult and "info" is undefined in destinationParachain:
+ * The transfer direction is from a Parachain to AH (from AT API)
+ *
+ * Default case return get transfer status for a transfer from Eth to a Parachain/AH
+ *
+ * @param transferResult - A transfer tracking response from the Snowbridge API or Subscan API.
+ * @returns - the transfer status
+ */
```
NunoAlexandre, line 64: Thanks for adding this 👌 👌 👌

NunoAlexandre (commented):
Looking very good overall :) a few comments but only the one about Sentry is more relevant imo

app/src/models/snowbridge.ts
Victor-Salomon: I initially wanted to do this, but I found out that namespaces are not a recommended approach nowadays

app/src/utils/snowbridge.ts
Victor-Salomon: We could indeed. I think the API route will notify sentry and return an empty array. 

Let me add this!

app/src/utils/snowbridge.ts
Victor-Salomon: Our registry does not have any subscan instances. 
But I agree it would be great to implement it. 

I also use them to create our explorer links. 
Let's create a linear task, and make a clean refacto that handles it everywhere.

NunoAlexandre: I was referring to the Asset Hub para id 👍

Victor-Salomon: Linear task [here](https://linear.app/velocitylabs/issue/VEL-96/handles-subscan-instances-url-in-registry-mainnet-and-testnet)

app/src/utils/transfer.ts
```diff
@@ -131,9 +131,8 @@ export function getExplorerLink(transfer: StoredTransfer): string | undefined {
       return `${removeURLSlash(explorersUrls.etherscan)}/address/${sender}`
     }
     case Network.Polkadot: {
-      if (result?.success?.assetHub && 'submittedAtHash' in result.success.assetHub)
-        return `${removeURLSlash(explorersUrls.subscan_assethub)}/block/${result.success.assetHub.submittedAtHash}`
-
+      if (result?.success?.assetHub && 'txHash' in result.success.assetHub)
```
noahjoeris, line 134: No blocker, but I want to emphasize that I'm usually a fan of abstracting something like this `'txHash' in result.success.assetHub)`. By creating a well named helper function that explains what its doing.

app/src/utils/transferTracking.ts
```diff
@@ -35,31 +34,80 @@ export const trackTransfers = async (
...
+  const isAhToEthTransfer =
+    'info' in transferResult && transferResult.info.destinationParachain == undefined
+
+  // Checks if the tracked XCM transfer comes from Subscan API.
+  const isXCMTransfer = 'destEventIndex' in transferResult && !('info' in transferResult)
+
+  if (isAhToEthTransfer)
+    return getTransferStatusFromParachain(transferResult as FromAhToEthTrackingResult)
+  if (isXCMTransfer)
+    return getTransferStatusFromParachain(transferResult as FromParachainTrackingResult)
```
noahjoeris, line 76: Amazing. That's very readable ❤️

app/src/utils/snowbridge.ts
Victor-Salomon: Ah, sorry, I understood the global block env.config 
Using the registry would require filtering on the environment which is not dynamic yet and can't be fetched from a hook...🥴
It could also be based on the new NEXT_PUBLIC_ENVIRONMENT. But its not as clean as it is to me.

noahjoeris (approved):
Great Victor. Also thanks for the efforts to document and produce clean code. Creating readable code is an art and you did an awesome job ❤️

app/src/utils/transfer.ts
Victor-Salomon: I've pushed a fix for this one. 
I remember both of you saying it, so I updated it yesterday :)

NunoAlexandre (approved):
🏄‍♂️

PR #149, opened 2024-10-14
⚡ improvement: registry logo urls

noahjoeris:
This small PR changes some registry logo urls to more reliable sources. 
And fixes the wrong WETH logo we used.


NunoAlexandre (approved):
💘

PR #155, opened 2024-10-23
fix: 🔍 get message_hash from UpwardMessageSent extrinsic

Victor-Salomon:
A little fix that should support finding the messageHash with x-tokens pallet.

### DIFF
--- a/app/src/utils/polkadot.ts
+++ b/app/src/utils/polkadot.ts
@@ -36,6 +36,11 @@ export const handleSubmittableEvents = (result: ISubmittableResult) => {
 
     // Filter the events to get the needed data
     events.forEach(({ event: { data, method, section } }) => {
+      // Supports x-tokens pallet
+      if (method === 'UpwardMessageSent' && section === 'parachainSystem') {
+        messageHash = data[0].toString()
+      }
+      // Supports XCM-polkadot pallet
       if (method === 'XcmpMessageSent' && section === 'xcmpQueue' && 'messageHash' in data) {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         messageHash = (data.messageHash as any).toString()
@@ -46,7 +51,6 @@ export const handleSubmittableEvents = (result: ISubmittableResult) => {
       }
     })
     if (!messageHash) throw new Error('Cross chain messageHash missing')
-    if (!messageId) throw new Error('Parachain messageId missing')
 
     return { messageHash, messageId }
   }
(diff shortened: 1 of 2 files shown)


app/src/utils/polkadot.ts
```diff
@@ -36,6 +36,11 @@ export const handleSubmittableEvents = (result: ISubmittableResult) => {
 
     // Filter the events to get the needed data
     events.forEach(({ event: { data, method, section } }) => {
+      // Supports x-tokens pallet
+      if (method === 'UpwardMessageSent' && section === 'parachainSystem') {
+        messageHash = data[0].toString()
+      }
+      // Supports XCM-polkadot pallet
```
NunoAlexandre, line 43: What was the transfer you tried to see this event? I'm asking because if I am not mistaken, `UpwardMessageSent` is when a transfer is destined to the Relay chain. So this will only work for that scenario, not for Relay -> other chain nor for Para -> Para

Victor-Salomon: @NunoAlexandre, thank you for the question!

Here is the link of a DOT transfer from Bifrost to Hydration: [Subscan](https://polkadot.subscan.io/xcm_message/polkadot-8c93739a3e683d77acd0c6300f46d8615c2f7a9d)

It's a proper **two-step transfer** as we can see. 
I'm eager to do more tests once we have some other routes supported in our registry.

NunoAlexandre: That makes sense:

The first step was the `UpwardMessageSent` as I thought, going to the Relay:

Screenshot 2024-10-23 at 16 41 09

And then from the Relay to Hydra:

Screenshot 2024-10-23 at 16 41 14

I guess that the event for Relay -> Hydra is `DownwardMessageSent`.

---

FYI, To cover Para to Para, you can try transferring GLMR from Moonbeam to Centrifuge for example 👍 That should use xTokens and show different events.

Victor-Salomon: @NunoAlexandre Amazing! I'll test this for sure. 
So I guess this 2 step-process through Relaychain is because of transferring DOT?

NunoAlexandre: Yes, because DOT is not a native token to neither Bifrost neither Hydration (or in other words here, neither Bifrost neither Hydration are Reserve chains of DOT). Imagine that you borrow me one of your books. For me to borrow it to Noah, I'd need to make the book pass through you first. That's the drill basically.

Victor-Salomon: Makes all sense. 
I'll keep you updated about the next tests.

NunoAlexandre (approved):
Ready to go ⛵️ can you just rename the title before merging to something more description? Merci bien 🥐

PR #156, opened 2024-10-24
feat ⚡ integrate paraspell xcm fee preview + add relay <-> AH route

noahjoeris:
This PR
- adds Paraspell xcm fee fetching
- adds debounce to fetching fees.
- adds DOT transfers from AH to relay and back
- refactors useFees a bit

Future improvements:
- possibly replace `getTransferInfo` with `getOriginFeeDetails` when available. Could be faster

### DIFF
--- a/app/src/hooks/useFees.tsx
+++ b/app/src/hooks/useFees.tsx
@@ -6,18 +6,12 @@ import { Token } from '@/models/token'
 import { Fees } from '@/models/transfer'
 import { getTokenPrice } from '@/services/balance'
 import { Direction, resolveDirection } from '@/services/transfer'
-import { getTokenSymbol } from '@/utils/paraspell'
+import { getChainNode, getTokenSymbol } from '@/utils/paraspell'
 import { toHuman } from '@/utils/transfer'
-import {
-  assets,
-  determineRelayChain,
-  getTransferInfo,
-  TNodeDotKsmWithRelayChains,
-} from '@paraspell/sdk'
+import { getTransferInfo } from '@paraspell/sdk'
 import { captureException } from '@sentry/nextjs'
 import { toEthereum, toPolkadot } from '@snowbridge/api'
 import { useCallback, useEffect, useRef, useState } from 'react'
-import useEnvironment from './useEnvironment'
 import useSnowbridgeContext from './useSnowbridgeContext'
 
 const DEBOUNCE_DELAY_MS = 500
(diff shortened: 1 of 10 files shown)


NunoAlexandre (approved):
Looks good :) nice job 🤩

app/src/utils/paraspell.ts
```diff
@@ -18,29 +26,66 @@ export const createTx = async (
...
-  return await Builder(api) // Api parameter is optional
-    .from(sourceChainFromId)
-    .to(destinationChainFromId)
-    .currency({ symbol: tokenSymbol })
-    /*.feeAsset(feeAsset) */
-    .amount(amount)
-    .address(recipient)
-    /*.xcmVersion(Version.V1/V2/V3/V4)*/
-    .build()
+  // from relay chain
```
Victor-Salomon, line 29: How about this ?

```  
const sourceChainFromId = assets.getTNode(sourceChain.chainId)
const destinationChainFromId = assets.getTNode(destinationChain.chainId)
if (!sourceChainFromId || !destinationChainFromId)
    throw new Error('Transfer failed: chain id not found.')

  // from relay chain
  if (sourceChain.chainId === 0)
    return await Builder(api).to(destinationChainFromId).amount(amount).address(recipient).build()

  // to relay chain
  else if (destinationChain.chainId === 0)
    return await Builder(api).from(sourceChainFromId).amount(amount).address(recipient).build()

  // para to para
  else {
    const tokenSymbol = getTokenSymbol(sourceChainFromId, token)
    return await Builder(api)
      .from(sourceChainFromId)
      .to(destinationChainFromId)
      .currency({ symbol: tokenSymbol })
      .amount(amount)
      .address(recipient)
      .build()
  }
  ```

File: app/src/hooks/useFees.tsx [hunk 2]
```diff
@@ -101,21 +111,34 @@ const useFees = (
...
+      fetchFees()
+    }, DEBOUNCE_DELAY_MS)
+
+    // Cleanup timeout on unmount
+    return () => {
+      if (debounceTimeoutRef.current) {
+        clearTimeout(debounceTimeoutRef.current)
+      }
+    }
+  }, [amount, fetchFees])
```
Victor-Salomon, line 141: Super nice 💥
I like this debounce! 
If somehow, we need a debounce anywhere else, it will be worth outsourcing it as a util 🤩

Victor-Salomon (approved):
Looks all good to me! 
Nice one again 🤩
I've left a comment :)

app/src/utils/paraspell.ts
noahjoeris: This will not work because assets.getTNode doesnt contain relay chain for some reason. So it returns undefined which throws an error in your case

Victor-Salomon: Oh ok.. not nice 🥲
So, it is all good for you to merge

PR #161, opened 2024-10-28
feat: Add Hydration

NunoAlexandre:
For now only enabling Ethereum -> Hydra transfers


app/src/config/registry.ts
```diff
@@ -390,6 +402,11 @@ export interface Registry {
   chains: Chain[]
   tokens: Token[]
   routes: Route[]
+  // Assets are often uniquely identified by a "asset id" at each chain, making it a chain-dependant value.
+  // The SDKs we use accept the token symbol as the indexing value to work with a given token but some chains
+  // have multiple tokens with the same symbol, in which case we need this map to provide the exact asset id
+  // to disambiguate. The map is Chain id -> Turtle token id -> Local Asset Id.
+  assetId: Map<string, Map<string, string>>
```
noahjoeris, line 409: What do you think of creating types here?
`Map<ChainId, Map<TokenId, LocalAssetId>>`
For me this is way cleaner and more readable.

app/src/config/registry.ts
```diff
@@ -547,3 +581,11 @@ export function rpcConnectionAsHttps(rpc?: string): string {
   if (!rpc) return ''
   return rpc.replace('wss://', 'https://')
 }
+
+export function getAssetId(env: Environment, chainId: string, tokenId: string): string | undefined {
+  return REGISTRY[env].assetId.get(chainId)?.get(tokenId)
+}
+
+export function isAssetHub(chain: Chain): boolean {
```
Victor-Salomon, line 589: I will see if I can use this one anywhere else 🤙

noahjoeris (approved):
Nice PR @NunoAlexandre ❤️

app/src/config/registry.ts
NunoAlexandre: It's aesthetically more pleasing but the types will just be aliases so there's no real added value imo. But we can add that if you like, albeit that it's already explained in the comment.

Open to it :)

Victor-Salomon (approved):
LGTM 💥🦾

app/src/config/registry.ts
NunoAlexandre: Done

PR #163, opened 2024-10-28
fix: update tracking status

Victor-Salomon:
- Add a condition to subscan query since finding transfer by extrinsic_id can only work for relaychain
- Switch subscan status definition from cross_chain_status param to status param
- Fix getTransferStatus helper

### DIFF
--- a/app/src/utils/subscan.ts
+++ b/app/src/utils/subscan.ts
@@ -24,7 +24,9 @@ export const trackFromParachainTx = async (
 
       const query = await relaychain.post('api/scan/xcm/list', {
         ...(crossChainMessageHash && { message_hash: crossChainMessageHash }),
-        ...(sourceChainExtrinsicIndex && { extrinsic_index: sourceChainExtrinsicIndex }),
+        // Only fetch by ExtrinsicIndex if sourceChain is Relay chain & crossChainMessageHash is undefined
+        ...(!crossChainMessageHash &&
+          sourceChainExtrinsicIndex && { extrinsic_index: sourceChainExtrinsicIndex }),
         row: 10,
       })
       if (query.status !== 200) {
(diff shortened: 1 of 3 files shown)


app/src/utils/transferTracking.ts
```diff
@@ -93,24 +93,19 @@ export function getTransferStatusFromParachain(
   const isBHChannelMsgDeliveredInSnowbridgeRes =
     'bridgeHubChannelDelivered' in transferResult &&
     transferResult.bridgeHubChannelDelivered?.success
-  /** Destination Event Index available */
-  const isDestEventIdxInSubscanXCMRes =
-    'destEventIndex' in transferResult && transferResult.destEventIndex.length > 0
+  /** Destination chain is Ethereum in XCM transfer*/
+  const isDestChainEThInSubscanXCMRes =
```
noahjoeris, line 97: typo

noahjoeris (approved):

app/src/utils/subscan.ts
```diff
@@ -42,18 +44,34 @@ export const trackFromParachainTx = async (
...
+      //     break
+      //   case 2:
+      //     xchainTransferStatus = TransferStatus.Failed
+      //     break
+      //   case 3:
+      //     xchainTransferStatus = TransferStatus.Complete
+      //     break
+      //   default:
+      //     break
+      // }
```
NunoAlexandre, line 61: in case you will want to uncomment and use this code, we should move this to a separate function 🙏

app/src/utils/subscan.ts
```diff
@@ -42,18 +44,34 @@ export const trackFromParachainTx = async (
...
+        case 'failed':
           xchainTransferStatus = TransferStatus.Failed
           break
-        case 3:
+        case 'success':
           xchainTransferStatus = TransferStatus.Complete
           break
+        // case ("relayed"):
+        //   xchainTransferStatus = TransferStatus.Complete
+        //   break
```
NunoAlexandre, line 72: out of curiosity, why is this commented? do you want to delete it?

app/src/utils/transferTracking.ts
NunoAlexandre, line 97: I read a small article the other day about not uppercasing in var names because it makes it hard to read and I got to agree  👀  and these names are also getting out of order, I wonder if we can somehow organise the code in a way that we don't have to put all of this info in the naming but understand a lot of from from the context (e.g. `isDestEthereum` instead of `isDestChainEThInSubscanXCMRes`. Just thinking out loud, not a blocker and not something for now)

```suggestion
  const isDestChainEthInSubscanXcmRes =
```

app/src/utils/subscan.ts
Victor-Salomon: Totally agreed :)
I'm expecting an answer from Yakio about this so I can integrate it.

app/src/utils/subscan.ts
Victor-Salomon: To be honest, I've never seen this case happening. 
I do not know if we should consider "relayed" as a completed 'transfer' since maybe nothing happens here once this "relayed" status is set. I will see once we tested all the flows/paths if we can get rid of this, or if Yakio can support understanding.

app/src/utils/transferTracking.ts
Victor-Salomon: I also feel naming is a bit messy. 
For now, I'm committing, your typo fix and once we finish tracking parachains transfers, I will clean and improve general namings, eventually outsourcing some functions, etc...

Victor-Salomon: => I've updated the naming. 
It makes more sense to do it now 🫠

app/src/utils/subscan.ts
NunoAlexandre: I see, yeah. I don't _think_ that `relayed` would mean `Complete`, I guess it's just meant to mark a message being bounced between a hop. Let's see what Yakio says 👍

PR #164, opened 2024-10-29
fix: remove unused papi code

noahjoeris:
- fixes rpc error [VEL-104](https://linear.app/velocitylabs/issue/VEL-104/bug-rpc-error-on-swap-chains-spam)


NunoAlexandre (approved):
🧹

PR #168, opened 2024-10-30
pre-commit: Add lint changes to commit

NunoAlexandre:
I personally find it quite annoying to have the pre-commit linting but not including the side effects in the commit itself.

We can either:
-  add them to the commit as I'm doing it, with the downside that we might be adding other untracked changes that might not be desired as part of the commit
- drop this pre-commit all together
- any other ideas?

I know that Rudolfs also doesn't quite enjoy the way this flows now so I am not alone on this one 😜

### DIFF
--- a/app/.husky/pre-commit
+++ b/app/.husky/pre-commit
@@ -1,6 +1,0 @@
-cd "$(dirname -- "$0")/.."
-
-pnpm run format
-pnpm run lint
-
-git add .
--- a/app/.husky/pre-commit
+++ b/app/.husky/pre-commit
@@ -2,3 +2,5 @@ cd "$(dirname -- "$0")/.."
 
 pnpm run format
 pnpm run lint
+
+git add .


noahjoeris:
Loool you were the one wanting to remove the `git add .` and complaining ahaha 🤣 
Therefore the only solution is to remove the hook then

noahjoeris:
But I will be the one kicking ass if lints warns are not removed

NunoAlexandre:
> Loool you were the one wanting to remove the `git add .` and complaining ahaha 🤣 Therefore the only solution is to remove the hook then

LOL really? 😆  probably because I suffered with the side effects.

I guess the best really would be to run `pnpm format` and `pnpm lint` from the code editor when saving changes, so that it's part of the workflow instead of a post-mortem step.

wdyt?

NunoAlexandre:
> feel free to remove the hook

@noahjoeris should we remove the whole `.husky` dir actually?

noahjoeris:
Tbh that's a very bold move. I can already see many issues arise. Unformatted code, Code from others is formatted, messed up PRs, accumulating lint warnings, noah can fix all these warnings 😆

noahjoeris:
So I'm personally against it but I go with the majority decision

noahjoeris:
I would actually prefer to research how to improve the hooks

Victor-Salomon:
haha! I was also part of the first complaint round 🤣
But because it was adding everything with the `git add .`

I'm not f&an of the flow either but I do not really care, I always merge the dest branch in mine before I open a PR and always format/lint/build before I commit 🫠

noahjoeris:
Ok let's say we remove it, but once we screw up 3 times we add it back again. I don't tolerate more than 3 errors 😝

NunoAlexandre:
Let's not lose time with this now, we can discuss a better approach when we meet up

PR #169, opened 2024-10-30
feat 💫 add new routes

noahjoeris:
Added routes 🚀 
- Moving DOT from RC to Hydration, Moonbeam, Interlay, Acala, Centrifuge, Astar, Bifrost
- Moving DOT from Hydration, Bifrost, Acala to RC
- routes are tested and working

Not working routes 🕵️ 
- Moving DOT from Moonbeam, Centrifuge, Astar to Relay. It fails to fetch balance, we need to investigate. Maybe local asset id will help.
- These are not included in this PR so we can do it in a seperate one


NunoAlexandre (approved):
BOOOM! 🐢 Is growing!!!

Victor-Salomon (approved):
Let's goooooo 🪄

PR #172, opened 2024-11-04
feat: add next batch of routes 

noahjoeris:


app/src/components/SubstrateWalletModal.tsx
```diff
@@ -33,8 +33,8 @@ const SubstrateWalletModal: FC = () => {
       open={isModalOpen}
       footer={isMobile && <WalletNotAccessible />}
       onWalletConnectClose={() => closeModal()}
-      onAccountSelected={account => setSubstrateAccount(account)}
-      onUpdatedAccounts={handleUpdatedAccounts}
+      onAccountSelected={account => setSubstrateAccount(account as Account)}
+      onUpdatedAccounts={acc => handleUpdatedAccounts(acc as Account[])}
```
NunoAlexandre, line 37: why this change?

app/src/config/registry.ts
```diff
@@ -262,6 +263,19 @@ export namespace Mainnet {
...
+
+  export const CFG: Token = {
+    id: 'cfg',
+    name: 'Centrifuge',
+    symbol: 'CFG',
+    logoURI: 'https://s2.coinmarketcap.com/static/img/coins/64x64/6748.png',
+    decimals: 18,
+    address: '',
+    multilocation:
+      '{"parents":"1","interior":{"X2":[{"Parachain":"2031"},{"GeneralKey":"0x0001"}]}}',
```
NunoAlexandre, line 275: my baby 👶  I gave it this multilocation by the end of 2021/early 2022, I feel so emotional <3

app/src/hooks/useFees.tsx
```diff
@@ -84,6 +84,12 @@ const useFees = (
             destinationChain,
           )
 
+          console.log('sourceChainNode', sourceChainNode)
+          console.log('destinationChainNode', destinationChainNode)
+          console.log('currency', currency)
+          console.log('amount', amount.toString())
+          console.log('senderAddress', senderAddress)
```
NunoAlexandre, line 91: Do we want to remove these before merging? I perosnally don't mind leaving some logs but just leaving this here as a reminder

app/src/utils/paraspell.ts
```diff
@@ -73,7 +73,7 @@ export const getTokenSymbol = (sourceChain: TNodeDotKsmWithRelayChains, token: T
   // TODO(victor): write some tests
   const supportedAssets = assets.getAllAssetsSymbols(sourceChain)
   const tokenSymbol = supportedAssets.find(a => a.toLowerCase() === token.symbol.toLowerCase())
-  if (!tokenSymbol) throw new Error('Transfer failed: Token symbol not supported.')
+  if (!tokenSymbol) throw new Error('Token symbol not supported.')
```
NunoAlexandre, line 76: I think this one is thrown as a notification to the user so we should reduce the technical language here. "Token not supported" is enough and clear for them.

```suggestion
  if (!tokenSymbol) throw new Error('Token not supported.')
```

NunoAlexandre (commented):
Looks good! so many new flows :) Just a few remarks

NunoAlexandre:
@noahjoeris can you add a description to this PR? The title is already vague and it would be nice for historical reasons to have more info in the description

app/src/hooks/useFees.tsx
Victor-Salomon: I would rather remove them :) ?

app/src/components/SubstrateWalletModal.tsx
Victor-Salomon: I guess casting `account as Account` now supports properly the Signer type.

Could we get this instead? 
`onUpdatedAccounts={accounts => handleUpdatedAccounts(accounts as Account[])}`

Victor-Salomon (approved):
I left one comment but all good mate !!

app/src/utils/paraspell.ts
noahjoeris: it is not shown to the user. So I keep it

NunoAlexandre (approved):
Biutifel!

PR #173, opened 2024-11-06
chore: paraspell version upgrade

Victor-Salomon:
Update Paraspell SDK to the latest version.
And fix useFees hook to make the build succeed.


noahjoeris (approved):
niceeee. looks great

PR #174, opened 2024-11-07
feat: Token origin badge

NunoAlexandre:
feat: Show a token origin badge

We display a badge for Ethereum-native tokens to differentiate them from potential overlapping Polkadot-native counterparts, like USDC/USDT.

When sending from Ethereum, we show it's an Ethereum token. But when that token is on a Polkadot chain, we show the origin badge as "Snowbridge" since that's the (only, for now) bridge that bridged and wrapped the token into the Polkadot ecosystem. In the future, we will support more bridges and we could have Ethereum USDC being Snowbridge USDC, Hyperbridge USDC, Axelar USDC, Wormhole USDC, etc. In that case, we should show the origin as the bridge that wrapped those tokens, since, for example, Axelar USDC is effectively a different token than Snowbridge USDC on Polkadot.

**Examples**
- When sourceChain is Ethereum and the token is wETH, we show `Ethereum` as the an origin badge
    
Screenshot 2024-11-07 at 13 55 27

- When sourceChain is AssetHub and the token is (snowbridge-wrapped) wETH, we show `Snowbridge` as the origin badge

Screenshot 2024-11-07 at 13 55 16


app/src/config/registry.ts
```diff
@@ -7,7 +7,7 @@ export namespace Mainnet {
   export const Ethereum: Chain = {
     uid: 'ethereum',
     name: 'Ethereum',
-    logoURI: 'https://s2.coinmarketcap.com/static/img/coins/64x64/1027.png',
+    logoURI: 'https://raw.githubusercontent.com/0xsquid/assets/main/images/webp128/tokens/eth.webp',
```
NunoAlexandre, line 10: We don't have to but I like this logo a lot better and there doesn't seem to be one single official logo so 🤷‍♂️

app/src/components/TokenLogo.tsx
```diff
@@ -0,0 +1,69 @@
...
+        alt={token.name}
+        width={32}
+        height={32}
+        className="token-logo h-[2rem] w-[2rem] rounded-full border-1 border-turtle-foreground bg-background"
+      />
+      {/* The origin label - either the origin chain or the bridge that has wrapped this token */}
+      {originBadge && (
+        <div className="absolute bottom-[-2px] right-[-1px] h-fit w-fit">
+          <div className="relative">
+            <Tooltip content={'Snowbridge'} showIcon={false}>
```
noahjoeris, line 30: Is snowbridge hardcoded here on purpose?

app/src/components/TokenLogo.tsx
```diff
@@ -0,0 +1,69 @@
...
+  logoURI: string
+  text: string
+}
+
+// Get the origin badge info for `token` given the context that it's being sent from `sourceChain`.
+// We don't display a badge if the token is native to the soureChain or it's network.
+// E.g:
+//   - When sourceChain is Ethereum and the token is wETH, we do *NOT* show an origin badge
+//   - When sourceChain is Hydration and the token is Polkadot-native USDC, we do *NOT* show an origin badge
+//   - When sourceChain is AsssetHub and the token is (snowbridge-wrapped) wETH, we *DO* show an origin badge
```
noahjoeris, line 56: Can you turn this into a doc comment? :)

app/src/models/token.ts
```diff
@@ -14,8 +14,15 @@ export interface Token {
...
   //   - https://www.coingecko.com/en/coins/ethereum
   coingeckoId?: string
+  // The origin of this token
+  origin: Origin
 }
 
+// The supported bridges. Each bridge will have it's own wrapped version of a token
+export type Bridge = 'Snowbridge'
+// The origin of a token
+export type Origin = { type: 'Ethereum'; bridge: Bridge } | { type: 'Polkadot'; paraId: number }
```
noahjoeris, line 24: Just a note: we also have the variables: `Network.Ethereum` and `Network.Polkadot`. But yeah I like the string values more as well. We are just mixing it now.

noahjoeris (approved):
Left some comments but nice job Nuno 🚀

noahjoeris:
I see it doesn't build. Let's also fix that before merging.

app/src/components/TokenLogo.tsx
NunoAlexandre: Good catch! Thanks

app/src/models/token.ts
NunoAlexandre: Yeah true. let me think about this

NunoAlexandre: @noahjoeris I am changing the `Network` enum to be a type `= 'Ethereum' | 'Polkadot'` because honestly it's a much nicer way of doing Enums in Typescript BUT for this `Origin` type, I can't reference that Network type as the type of `Origin.type` so I need to leave it like this

NunoAlexandre: Unless if you know some magic trick 👁️

NunoAlexandre:
@noahjoeris all addressed and build fixed. I want to test it before merging since there are implications in the schema that affects local storage.

app/src/models/token.ts
Victor-Salomon: To use the `Network` enum, I only see generics as a solution, but it would add more complexity for no benefits

NunoAlexandre: yeah I agree @Victor-Salomon 🙏

Victor-Salomon (approved):
LGTM 🧨

NunoAlexandre:
> I see the lint/prettier check failing. Can you please check

it's already fixed

app/src/components/TokenLogo.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/**
+ *  Get the origin badge info for `token` given the context that it's being sent from `sourceChain`.
+ *  We don't display a badge if the token is native to the soureChain or it's network.
+ *  E.g:
+ *    - When sourceChain is Ethereum and the token is wETH, we do *NOT* show an origin badge
+ *   - When sourceChain is Hydration and the token is Polkadot-native USDC, we do *NOT* show an origin badge
+ *    - When sourceChain is AsssetHub and the token is (snowbridge-wrapped) wETH, we *DO* show an origin badge
+ */
```
noahjoeris, line 58: I don't mind but this is a crazy doc comment 🤣 usually you have one header and a @remarks section

noahjoeris (commented):
Looks good 👍🏻👍🏻👍🏻 just the lint/prettier check

noahjoeris:
Perfect 😃

Victor-Salomon (approved):
🦾

PR #175, opened 2024-11-07
feat: bifrost xcm

Victor-Salomon:
Quick new routes added for Bifrost parachain. 
Also, it improves Subscan error readability.

Bifrost parachain to be updated with USDC, USDT, USDC.e, USDT.e


app/src/utils/subscan.ts
```diff
@@ -41,6 +41,7 @@ export const trackXcm = async (
         console.log(
           `XCM transfer not found for ${crossChainMessageHash ? `message hash: ${crossChainMessageHash}` : `extrinsic id: ${sourceChainExtrinsicIndex}`}.`,
         )
+        query.json?.message && console.log(`Additional information: ${query.json.message}`)
```
noahjoeris, line 44: Do we wanna keep the logs?

app/src/utils/subscan.ts
```diff
@@ -59,6 +60,7 @@ export const trackXcm = async (
         const uniqueIdData: SubscanTransferResponse = uniqueIdQuery.json?.data
         if (!uniqueIdData) {
           console.log(`XCM transfer not found for unique_id: ${unique_id}`)
+          query.json?.message && console.log(`Additional information: ${query.json.message}`)
```
noahjoeris, line 63: here as well

noahjoeris (approved):
👍

app/src/utils/subscan.ts
Victor-Salomon: Yes, we now need them. 
This comment is only displayed to improve error understanding when the query is successful but the request can't be proceeded like yesterday with the requested payment.

NunoAlexandre (approved):
👌🏻👌🏻👌🏻

PR #176, opened 2024-11-08
fix: Lookup AH Local AssetId when transferring to AH

NunoAlexandre:
Currently when handling a transfer aimed at AssetHub, we are just looking up the token symbol there but we should lookup the local asset uid instead.

```
// current
const destinationChainNode = assets.getTNode(destinationChain.chainId, relay)
if (!destinationChainNode) throw new Error('Transfer failed: chain id not found.')
return { symbol: getTokenSymbol(destinationChainNode, token) }
```

Fixed that and cleaned up the code a little


NunoAlexandre:
@Victor-Salomon @noahjoeris can you please review this? 🙏

app/src/config/registry.ts
```diff
@@ -793,6 +793,12 @@ export const mainnetRegistry: Registry = {
       sdk: 'ParaSpellApi',
       tokens: [Mainnet.DOT.id],
     },
+    {
```
NunoAlexandre, line 796: Moved this here so it's after another Hydra route

Victor-Salomon (approved):
LGTM 💥

PR #178, opened 2024-11-12
feat: add Native USDC and USDT stablecoins

Victor-Salomon:
Add both Native USDC and USDT stablecoins. 
Tested on AH <=> Bifrost.


NunoAlexandre (approved):
👌

PR #180, opened 2024-11-12
fix: Restructure registry

NunoAlexandre:
### Changes

- Move registry from `./config/registry` to `./registry` (as its own module)
- Move mainnet and testnet configs to their own registry-submodule
- Namespace tokens following to their origin chain to avoid name overlapping
- Other small improvements to DRY repeated code


app/src/config/index.tsx
```diff
@@ -31,3 +31,5 @@ export const config = defaultWagmiConfig({
   }),
   // ...wagmiOptions, // Optional - Override createConfig parameters
 })
+
+export const DWELLIR_KEY = process.env.NEXT_PUBLIC_DWELLIR_KEY
```
Victor-Salomon, line 35: Let's remove the one from `config>registry.ts`
image

Victor-Salomon (approved):
Nice job mate 🤩

app/src/registry/index.ts
```diff
@@ -0,0 +1,99 @@
...
+    bridge: 'Snowbridge',
+  }
+}
+
+// Hack - importing this after types definition to avoid circular dependency and import timing issues
+import * as Mainnet from './mainnet'
+export * as Mainnet from './mainnet'
+
+import * as Testnet from './testnet'
+export * as Testnet from './testnet'
```
noahjoeris, line 99: Can you explain why we need that?

noahjoeris (approved):
Good job as always 💯

app/src/registry/index.ts
NunoAlexandre: So that we can use this inner module outside of `registry` without needing to wildcard-export everything the inner module (in this case `testnet` contains). In other words, we can do `registry.Testnet.<xyz>` instead of `registry.<xyz>`, granted that most likely `xyz` would overlap between mainnet and testnet and even if that wouldn;t be the case it wouldn't be clear the environment of an object we would be referencing.

PR #182, opened 2024-11-18
fix: eth network balance fetching

noahjoeris:
- fixes the error of unsuccessfully fetching token balances on ethereum network


Victor-Salomon (approved):
nice and simple fix 😎

app/src/registry/index.ts
```diff
@@ -93,7 +95,6 @@ export function snowbridgeWrapped(): Origin {
 
 // Hack - importing this after types definition to avoid circular dependency and import timing issues
 import * as Mainnet from './mainnet'
-export * as Mainnet from './mainnet'
```
NunoAlexandre, line 96: Can we undo this change?

app/src/registry/index.ts
```diff
@@ -46,6 +46,8 @@ export const SNOWBRIDGE_MAINNET_PARACHAIN_URLS = [
 ]
 
 export function getNativeToken(chain: Chain): Token {
+  if (chain.network === 'Ethereum') return Mainnet.Eth.ETH
```
NunoAlexandre, line 49: 👌🏻

app/src/registry/index.ts
NunoAlexandre: What command do you run? I'm running lint and format and it doesn't do that and frankly I don't quite like the change because it moves the export away :/ but it's not important

NunoAlexandre (approved):
🙌🏻

PR #184, opened 2024-11-18
fix: use private rpc endpoints with paraspell

noahjoeris:
This PR shifts from using public nodes to private nodes when making paraspell calls. 
I changed it for balance, fees and tx calls.


app/src/hooks/useFees.tsx
```diff
@@ -1,9 +1,9 @@
-import { getNativeToken } from '@/registry'
 import useNotification from '@/hooks/useNotification'
 import { Chain } from '@/models/chain'
 import { NotificationSeverity } from '@/models/notification'
 import { getCoingekoId, Token } from '@/models/token'
 import { Fees } from '@/models/transfer'
+import { getNativeToken } from '@/registry'
```
NunoAlexandre, line 6: Do we need this import here?

NunoAlexandre (approved):
Easy and effective 🙌🏻 there's a merge conflict, let me know if help is needed

app/src/hooks/useFees.tsx
noahjoeris: It is gone after the merge conflict resolve

noahjoeris: nope build failed without it. Will add it again.

NunoAlexandre (approved):
LG!!

PR #185, opened 2024-11-18
feat: Log transactions to Firebase

rudolfs:
This logs txs to a Firestore db. In another step we could write a stored function that periodically goes over the data and writes an aggregate somewhere that could be read out by the website and displayed.

Screenshot 2024-11-18 at 14 40 03
Screenshot 2024-11-18 at 14 40 13

Note: there's no authentication in the front-end, so potentially anyone could write to this db. The only restriction at the moment is that only creation is allowed, updates and deletes not. Current ruleset:
```
rules_version = '2';

service cloud.firestore {
  match /databases/{database}/documents {
    match /{document=**} {
      allow create: if true;
      allow update, delete: if false;
    }
  }
}
```


app/src/hooks/useSnowbridgeApi.tsx
```diff
@@ -181,19 +180,19 @@ const useSnowbridgeApi = () => {
...
-          destinationChain: destinationChain.name,
-          usdValue: tokenUSDValue ?? 0,
-          usdFees: fees.inDollars,
-          recipient: recipient,
-          date: date.toISOString(),
-        })
+      trackTransferMetrics({
+        id: sendResult.success?.messageId,
+        environment,
+        sender: senderAddress,
```
NunoAlexandre, line 186: 👌

app/src/utils/analytics.ts
```diff
@@ -11,58 +12,24 @@ export interface TransferMetric {
...
-      token: data.token,
-    },
-    unit: 'USD',
-  })
-
-  // Track unique senders (set)
-  metrics.set('unique_senders', data.sender)
+      usdFees: data.usdFees,
+      usdValue: data.usdValue,
+      environment: data.environment,
```
NunoAlexandre, line 30: @rudolfs to confirm, this means we now also trakc metrics for testnet transfers, right? I am not sure we want that. Let's keep the filter for Mainnet-only for now 🙏

app/src/utils/analytics.ts
```diff
@@ -1,7 +1,8 @@
-import { metrics } from '@sentry/nextjs'
-import { track } from '@vercel/analytics/react'
+import { db, setDoc, doc } from '@/config/firebase'
```
NunoAlexandre, line 1: NIT but wouldn't mind if we import it like this:

```suggestion
import * as Firebase from '@/config/firebase'
```

So that the call sites are more contextual and informative without needing to click around to understand what's happening.

NunoAlexandre (commented):
great to see this coming along 🚀

app/src/utils/analytics.ts
rudolfs: Correct. I can add the if statement back, ez. Just thought testing would be easier, but then I realized that we test on mainnet anyway, unless I misunderstood something?

rudolfs: Fixed in 001a60f74d076b2f3f26781968876633bfb0960a.

app/src/utils/analytics.ts
rudolfs: Fixed in 001a60f74d076b2f3f26781968876633bfb0960a. Why do have a lint rule that specifically disallows this though if it's encouraged in the codebase?

noahjoeris: I never encouraged this 👀 Nuno is taking over right now haha

noahjoeris: In fact, I am strongly against this :)

app/src/utils/analytics.ts
```diff
@@ -13,56 +14,23 @@ export interface TransferMetric {
...
+        sender: data.sender,
+        sourceChain: data.sourceChain,
+        token: data.token,
+        usdFees: data.usdFees,
+        usdValue: data.usdValue,
+      },
+    )
+  } catch (error) {
+    console.error('Error, was not able to log transaction to Firestore: ', error)
+  }
```
noahjoeris, line 35:
```suggestion
  } catch (error) {
    console.error('Error, was not able to log transaction to Firestore: ', error)
    captureException(error)
  }
```
and adding the sentry import

app/src/utils/analytics.ts
NunoAlexandre: Yeah that's correct :) Thanks!

app/src/utils/analytics.ts
NunoAlexandre: @noahjoeris can you explain why you are strongly against this?

app/src/utils/analytics.ts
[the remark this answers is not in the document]
rudolfs: 4567e6af8fa0bcece3e696485b706057a883d186

app/src/utils/analytics.ts
rudolfs: Tbh, I also prefer explicit imports unless there are namespace clashes or a lot of stuff to import. Tree-shaking is one benefit, IDEs adding explicit imports via hot-key is another.

fa1de5900344f95406e4aaece37fb59a7d09170f

NunoAlexandre: Yeah that's fine, let's go with that 🙌🏻

app/package.json
```diff
@@ -41,6 +41,7 @@
     "clsx": "^2.1.1",
     "cypress-wait-until": "^3.0.2",
     "ethers": "^6.13.4",
+    "firebase": "^11.0.2",
```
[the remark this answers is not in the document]
rudolfs: So this works and we don't need the dependency: c3033bb12d90b0eec8cb1c9610dc374996e2a449.

NunoAlexandre (approved):
awesome 👌

noahjoeris (approved):
Great Rudolfs 🚀

NunoAlexandre:
> Great Rudolfs 🚀

Rudolfs, The Great 🙌🏻

rudolfs:
> > Great Rudolfs 🚀
> 
> Rudolfs, The Great 🙌🏻

lol : D

PR #186, opened 2024-11-19
chore: Update to Paraspell 7.2.0

NunoAlexandre:
Minimal changes, basically just bumping the dependency and dropping the edge case for AH as a destination chain.

It fixes loading feed and submitting txs destined to Asset Hub.


noahjoeris (approved):
good

PR #188, opened 2024-11-19
fix: cache token price

Victor-Salomon:
Implemented solution: 

_refactor `useTokenPrice` hook_

- reactQuery + client/user browser cache 
- getApiTokenPrice handler to fetch data from API
- new API to cache the token price globally

@NunoAlexandre To be done next in a dedicated PR: 
Refacto the moment where `TxSummary` component consumes/fetch the `tokenPrice` API. I would rather avoid having the call being made in the `useTransferForm` and passed as a prop but directly called in the `TxSummary` component when needed. 

We can also adapt in seconds the useTokenPrice hook to provide a loading state if needed.


app/src/services/balance.ts
```diff
@@ -56,3 +25,19 @@ export const getTokenPrice = async (tokenId: string): Promise<TokenPrice | null>
     return null
   }
 }
+
+export const getApiTokenPrice = async (token: Token) => {
```
NunoAlexandre, line 29: Can you leave a docstring here explaining this basically wraps the `getTokenPrice` function above in a route call that's cached and so on? 🙏

app/src/services/balance.ts
```diff
@@ -56,3 +25,19 @@ export const getTokenPrice = async (tokenId: string): Promise<TokenPrice | null>
...
+  const response = await fetch(`/api/tokenPrice`, {
```
NunoAlexandre, line 30: I might be wrong but don't REST apis follow the lowercase + dash convention? so that would make this `/api/token-price`.

NunoAlexandre (commented):
This is a huge improvement and the changes look good! 

Just one remark: the API should return the `TokenPrice` type and not default to `0` if it failed to fetch the price. For two reasons:

1. So that the caller of the api can handle that scenario whatever best way they find 
2. Shall we need to offer the price in another currency that's not usd, we don't need to change the api

app/src/services/balance.ts
noahjoeris, line 29: Can we rename this? What about `getCachedTokenPrice`

app/src/services/balance.ts
noahjoeris: agree

noahjoeris (approved):

app/src/app/api/tokenPrice/route.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    const { token } = requestValue.data
+    if (!token) return NextResponse.json({ error: 'Token is undefined in API' }, { status: 400 })
+
+    const fetchTokenPrice = unstable_cache(
+      async () => {
+        const result = await getTokenPrice(getCoingekoId(token))
+        return result?.usd ?? 0
+      },
+      [`tokenPrice-${token.id}`],
```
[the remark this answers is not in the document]
Victor-Salomon: Yes it is 👍

app/src/services/balance.ts
Victor-Salomon: ✅

Victor-Salomon:
@noahjoeris @NunoAlexandre 
Thanks for the nice feedback!

@NunoAlexandre I've updated the return type of the API to make it more generic as requested.

PR #191, opened 2024-11-21
Improvement: get token price

Victor-Salomon:
Implement cached token price fetch in: 

- useFees
- useSnowbridgeApi
- useParaspellApi

And support allowanceLoading checks before displaying the TxSummy component to avoid overlaps. 

Every call to goingecko API is now cached globally.

### DIFF
--- a/app/src/components/Transfer.tsx
+++ b/app/src/components/Transfer.tsx
@@ -124,6 +124,14 @@ const Transfer: FC = () => {
     !requiresErc20SpendApproval &&
     !loadingFees
 
+  const shouldDisplayTxSummary =
+    isValid &&
+    tokenAmount &&
+    tokenAmount.token &&
+    !!tokenAmount.amount &&
+    !allowanceLoading &&
+    !requiresErc20SpendApproval
+
   return (
     <form
       onSubmit={handleSubmit}
@@ -317,19 +325,14 @@ const Transfer: FC = () => {
         )}
       </AnimatePresence>
 
-      {tokenAmount &&
-        tokenAmount.token &&
-        !!tokenAmount.amount &&
-        !allowanceLoading &&
-        !requiresErc20SpendApproval && (
-          <TxSummary
-            hidden={!isValid || requiresErc20SpendApproval}
-            loading={loadingFees || !fees}
-            tokenAmount={tokenAmount}
-            fees={fees}
-            durationEstimate={durationEstimate}
-          />
-        )}
+      {shouldDisplayTxSummary && (
+        <TxSummary
+          loading={loadingFees || !fees}
+          tokenAmount={tokenAmount}
+          fees={fees}
+          durationEstimate={durationEstimate}
+        />
+      )}
 
       {/* Transfer Button */}
       <Button
--- a/app/src/components/Transfer.tsx
+++ b/app/src/components/Transfer.tsx
@@ -61,6 +61,7 @@ const Transfer: FC = () => {
 
   const {
     allowance: erc20SpendAllowance,
+    loading: allowanceLoading,
     approveAllowance,
     approving: isApprovingErc20Spend,
   } = useErc20Allowance({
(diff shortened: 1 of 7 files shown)


app/src/components/Transfer.tsx
```diff
@@ -316,15 +317,19 @@ const Transfer: FC = () => {
...
-      {tokenAmount && tokenAmount.token && !!tokenAmount.amount && !requiresErc20SpendApproval && (
-        <TxSummary
-          hidden={!isValid || requiresErc20SpendApproval}
-          loading={loadingFees || !fees}
-          tokenAmount={tokenAmount}
-          fees={fees}
-          durationEstimate={durationEstimate}
-        />
-      )}
+      {tokenAmount &&
```
NunoAlexandre, line 320: Let's create a const "shouldDisplayTxSummary" to clean this up here

app/src/components/TxSummary.tsx
```diff
@@ -26,7 +26,7 @@ const TxSummary: FC<TxSummaryProps> = ({
   durationEstimate,
   hidden,
 }) => {
-  const { price, isTokenPriceLoading } = useTokenPrice(tokenAmount.token)
+  const { price, loading: isTokenPriceLoading } = useTokenPrice(tokenAmount.token)
```
NunoAlexandre, line 29: A more natural semantic here would be "isLoadingTokenPrice"

app/src/hooks/useFees.tsx
```diff
@@ -47,15 +47,19 @@ const useFees = (
         case Direction.ToEthereum: {
           if (!snowbridgeContext) throw new Error('Snowbridge context undefined')
 
-          tokenUSDValue = (await getTokenPrice('polkadot'))?.usd ?? 0
+          const polkadot = REGISTRY[env].tokens.find(t => t.name === 'Polkadot')
```
NunoAlexandre, line 50: Let's not do this. We are adding a dependency and a O(n) lookup on the registry that filters on a hardcoded value just to pass the coingeko Id for DOT which we statically know to be "polkadot" so let's just pass that directly instead as it was. A compromise is to reference 'mainnet.Polkadot.DOT'.

app/src/hooks/useFees.tsx
```diff
@@ -47,15 +47,19 @@ const useFees = (
...
+
           fees = (await toEthereum.getSendFee(snowbridgeContext)).toString()
           break
         }
 
         case Direction.ToPolkadot: {
           if (!snowbridgeContext) throw new Error('Snowbridge context undefined')
 
-          tokenUSDValue = (await getTokenPrice('ethereum'))?.usd ?? 0
+          const ethereum = REGISTRY[env].tokens.find(t => t.name === 'Ethereum')
```
NunoAlexandre, line 60: Same here about Ethereum

app/src/registry/mainnet.ts
```diff
@@ -468,6 +468,7 @@ export const REGISTRY: Registry = {
   ],
   tokens: [
     Eth.WETH,
```
NunoAlexandre, line 470: We should remove this. I understand you needed to add it to filter the registry to find it and get the coingeko Id but read my comment above about undoing that for ETH and DOT 🙏

NunoAlexandre (commented):
Nice stuff :) just a few comments

app/src/hooks/useFees.tsx
Victor-Salomon: @NunoAlexandre Al right! 
Since, `getCachedTokenPrice()` expects a `token:Token` as param, and not the `tokenId:string`, what do you think about this? 

=> `tokenUSDValue = (await getCachedTokenPrice(Eth.ETH))?.usd ?? 0`
=> `tokenUSDValue = (await getCachedTokenPrice(Polkadot.DOT))?.usd ?? 0`

It still relies on the registry but directly on the token itself, no more lookup, and we can still update according to the `env` (mainnet/testnet) later. 

Otherwise, I need refactoring both the API handler and the API itself directly.

NunoAlexandre: That's good, that's what I meant by the "compromise" option 👍  Let's go :)

app/src/components/Transfer.tsx
noahjoeris: 100% agree

app/src/registry/mainnet.ts
noahjoeris: agree

noahjoeris (changes requested):
What nuno said

NunoAlexandre (approved):
image

noahjoeris (approved):

PR #192, opened 2024-11-25
fix: tx summary displaying flow

Victor-Salomon:
Fix flow when the form is valid but Tx Summary is not displayed.


NunoAlexandre (approved):
Awesome :) 🚀

PR #194, opened 2024-11-25
ux: Improve XCM tx send flow

NunoAlexandre:
### Main changes

- Morph the `SendButton` to show status updates
- When sending an XCM, move the tx to 'ongoing' as soon as possible 
- Lots of (hopefully) improvements

### DIFF
--- a/app/src/components/Button.tsx
+++ b/app/src/components/Button.tsx
@@ -16,7 +16,7 @@ const styles = {
     'border-1 border-turtle-level3 bg-transparent hover:border-turtle-level3 disabled:border-turtle-level3 disabled:opacity-30',
   ghost: 'bg-transparent disabled:opacity-30',
   update:
-    'border-1 border-turtle-level3 bg-transparent disabled:border-turtle-secondary text-turtle-secondary-dark',
+    'border-1 border-turtle-secondary-dark bg-turtle-secondary text-white disabled:opacity-100 disabled:bg-opacity-50 disabled:border-opacity-40',
 }
 
 const sizeHeights: Record<ButtonSize, string> = {
@@ -82,6 +82,7 @@ const Button: FC<ButtonProps> = ({
       onClick={onClick}
       size={size}
       radius="sm"
+      disableAnimation={true}
       className={twMerge('w-full', styles[variant], sizeHeights[size], paddingX[size], className)}
       type={type}
       style={{
--- a/app/src/components/Button.tsx
+++ b/app/src/components/Button.tsx
@@ -3,6 +3,7 @@ import { Button as NextButton } from '@nextui-org/react'
 import { FC, ReactNode } from 'react'
 import { twMerge } from 'tailwind-merge'
 import LoadingIcon from './svg/LoadingIcon'
+import { TransferStatus } from '@snowbridge/api/dist/history'
 
 type ButtonVariant = 'primary' | 'secondary' | 'outline' | 'ghost'
 type ButtonSize = 'sm' | 'md' | 'lg'
@@ -55,6 +56,7 @@ interface ButtonProps {
   type?: 'button' | 'submit' | 'reset'
   /** Cypress ID for testing. */
   cypressID?: string
+  status: TransferStatus
 }
 
 const Button: FC<ButtonProps> = ({
@@ -68,6 +70,7 @@ const Button: FC<ButtonProps> = ({
   children,
   type = 'button', // Default type is 'button',
   cypressID,
+  status,
 }) => {
   return (
     <NextButton
@@ -85,11 +88,13 @@ const Button: FC<ButtonProps> = ({
     >
       {/** loading state */}
       {loading && (
-        <LoadingIcon
-          className="animate-spin"
+        <div className='flex items-center'><LoadingIcon
+          className="animate-spin mr-3"
           width={spinnerSize[size]}
           height={spinnerSize[size]}
         />
+        {status}
+        </div>
       )}
 
       {/** children or label */}
--- a/app/src/components/Button.tsx
+++ b/app/src/components/Button.tsx
@@ -16,7 +16,7 @@ const styles = {
     'border-1 border-turtle-level3 bg-transparent hover:border-turtle-level3 disabled:border-turtle-level3 disabled:opacity-30',
   ghost: 'bg-transparent disabled:opacity-30',
   update:
-    'border-1 bg-turtle-secondary border border-turtle-secondary-dark text-white disabled:opacity-100 disabled:bg-opacity-50 disabled:border-opacity-40',
+    'border-1 bg-turtle-secondary border border-turtle-secondary-dark text-turtle-secondary-dark disabled:opacity-100 disabled:bg-opacity-50 disabled:border-opacity-40',
 }
 
 const sizeHeights: Record<ButtonSize, string> = {
--- a/app/src/components/Button.tsx
+++ b/app/src/components/Button.tsx
@@ -4,7 +4,7 @@ import { FC, ReactNode } from 'react'
 import { twMerge } from 'tailwind-merge'
 import LoadingIcon from './svg/LoadingIcon'
 
-type ButtonVariant = 'primary' | 'secondary' | 'outline' | 'ghost'
+type ButtonVariant = 'primary' | 'secondary' | 'outline' | 'ghost' | 'update'
 type ButtonSize = 'sm' | 'md' | 'lg'
 
 const styles = {
@@ -15,6 +15,8 @@ const styles = {
   outline:
     'border-1 border-turtle-level3 bg-transparent hover:border-turtle-level3 disabled:border-turtle-level3 disabled:opacity-30',
   ghost: 'bg-transparent disabled:opacity-30',
+  update:
+    'border-1 border-turtle-level3 bg-transparent disabled:border-turtle-secondary text-turtle-secondary-dark',
 }
 
 const sizeHeights: Record<ButtonSize, string> = {
@@ -46,6 +48,8 @@ export interface ButtonProps {
   disabled?: boolean
   /** Whether the button is in a loading state. */
   loading?: boolean
+  /** A custom icon to be displayed instead of the spinner */
+  icon?: ReactNode
   /** The variant determines the preset color and style of the button. */
   variant?: ButtonVariant
   /** The size of the button. */
@@ -64,6 +68,7 @@ const Button: FC<ButtonProps> = ({
   className,
   disabled,
   loading,
+  icon,
   variant = 'primary',
   size = 'lg',
   children,
@@ -77,7 +82,7 @@ const Button: FC<ButtonProps> = ({
       onClick={onClick}
       size={size}
       radius="sm"
-      className={twMerge(styles[variant], sizeHeights[size], paddingX[size], className)}
+      className={twMerge('w-full', styles[variant], sizeHeights[size], paddingX[size], className)}
       type={type}
       style={{
         outline: 'none',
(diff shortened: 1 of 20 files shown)


app/src/components/Button.tsx
```diff
@@ -66,7 +71,14 @@ const Button: FC<ButtonProps> = ({
   variant = 'primary',
   size = 'lg',
   children,
-  type = 'button', // Default type is 'button',
+  icon = (
+    <LoadingIcon
+      className="mr-3 animate-spin"
+      width={spinnerSize[size]}
+      height={spinnerSize[size]}
+    />
```
NunoAlexandre, line 79: Worth checking if we can style the button svg color differently based on the button variant, to avoid us having to pass it explicitly with the desired color

app/src/components/Menu.tsx
```diff
@@ -8,14 +8,14 @@ const Menu = ({
   setNewTransferInit,
   hasCompletedTransfers,
 }: DisplaysTransfers) => {
-  const initNewTransaction = newTransferInit === TransferTab.New
+  const initNewTransaction = newTransferInit === 'New'
```
NunoAlexandre, line 11: I should have done these changes in a separate PR, my bad.

NunoAlexandre: I was trying to move the ongoing txs to a new 'Ongoing' tab but I didn't like the looks of it but kept these changes moving from enum to type.

app/src/components/SendButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const [{ label, icon, variant }, setProps] = useState<Props>(getProps(status, props.label))
+
+  useEffect(() => {
+    setProps(getProps(status, props.label))
+  }, [status, disabled])
+
+  return (
+    <div className="w-full">
+      <Button {...props} variant={variant} disabled={disabled} label={label} icon={icon} />
+    </div>
```
NunoAlexandre, line 19: Simplify

app/src/hooks/useParaspellApi.tsx
```diff
@@ -33,77 +34,107 @@ const useParaspellApi = () => {
...
-            })
+          try {
+            const eventsData = handleSubmittableEvents(result)
+            if (eventsData) {
+              const { messageHash, messageId, extrinsicIndex } = eventsData
 
-            // metrics
-            if (environment === Environment.Mainnet) {
-              trackTransferMetrics({
+              addOngoingTx({
```
NunoAlexandre, line 83: Document

app/src/hooks/useTransferForm.tsx
```diff
@@ -190,8 +193,17 @@ const useTransferForm = () => {
         recipient: recipient,
         fees,
         onSuccess: () => {
```
NunoAlexandre, line 195: `onSuccess` is a misleading name

app/src/models/transfer.ts
```diff
@@ -86,10 +86,7 @@ export interface AmountInfo {
   inDollars: number
 }
 
-export enum TransferTab {
-  New = 'New',
-  Completed = 'Completed',
-}
+export type TransferTab = 'New' | 'Ongoing' | 'Done'
```
NunoAlexandre, line 89:
```suggestion
export type TransferTab = 'New' | 'Done'
```

app/tailwind.config.ts
```diff
@@ -91,7 +91,12 @@ const config: Config = {
     },
   },
   darkMode: 'class',
-  plugins: [addVariablesForColors, nextui(), require('tailwindcss-animate')],
+  plugins: [
+    addVariablesForColors,
+    nextui(),
+    require('tailwindcss-animate'),
+    require('tailwindcss-motion'),
```
NunoAlexandre, line 98: We should probably remove this lib, not worth the extra weight for now I'd say

app/src/components/SendButton.tsx
```diff
@@ -0,0 +1,85 @@
+import { FC, ReactNode, useEffect, useState } from 'react'
+import { Status } from '@/hooks/useTransfer'
+import Button, { ButtonProps, ButtonVariant } from './Button'
+import LoadingIcon from './svg/LoadingIcon'
+import { colors } from '../../tailwind.config'
+
+type SendButtonProps = ButtonProps & { status: Status }
+
+const SendButton: FC<SendButtonProps> = ({ status, disabled, ...props }) => {
+  const [{ label, icon, variant }, setProps] = useState<Props>(getProps(status, props.label))
```
noahjoeris, line 10: Hmm. Props is usually the name for Component parameters only. I have never seen it in useState. Is it possible to rename it? Also an opportunity to be more precise.

noahjoeris: Ahh I see. You are storing the props of the other button component.

app/src/components/SendButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+  }, [status, disabled])
+
+  return <Button {...props} variant={variant} disabled={disabled} label={label} icon={icon} />
+}
+
+interface Props {
+  label?: string
+  icon?: ReactNode
+  variant?: ButtonVariant
+}
```
noahjoeris, line 23: We can use the interface of the `Button `component. No need to recreate it here. Maybe also worth thinking about renaming the component from `Button `to `BaseButton ` or similar

app/src/components/SendButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+function getProps(status: Status, defaultLabel?: string): Props {
```
noahjoeris, line 25:
```suggestion
function getButtonProps(status: Status, defaultLabel?: string): Props {
```

app/src/hooks/useParaspellApi.tsx
```diff
@@ -11,10 +11,11 @@ import { txWasCancelled } from '@/utils/transfer'
 import { captureException } from '@sentry/nextjs'
 import useNotification from './useNotification'
 import useOngoingTransfers from './useOngoingTransfers'
-import { Status, TransferParams } from './useTransfer'
+import { Sender, Status, TransferParams } from './useTransfer'
+import { ISubmittableResult } from '@polkadot/types/types'
 
 const useParaspellApi = () => {
-  const { addTransfer: addTransferToStorage } = useOngoingTransfers()
+  const { addOrUpdate } = useOngoingTransfers()
```
noahjoeris, line 18: maybe rename?
`addOrUpdateOngoingTransfers`

app/src/components/SendButton.tsx
NunoAlexandre: I actually renamed it already to `OverwrittenProps` before this comment

app/src/components/SendButton.tsx
NunoAlexandre: > We can use the interface of the Button component

What do you suggest? We could use the `Exclude` approach but that seems more of an hassle that a benefit for example

app/src/components/SendButton.tsx
NunoAlexandre: I don't like this. This is a local function in the `SendButton` props so there's no need to add "Button" again to the naming here imo

app/tailwind.config.ts
noahjoeris: which one? The tailwind motion? What did you use it for?

app/src/hooks/useParaspellApi.tsx
NunoAlexandre: I did that first but I find it's unnecessarily too long. Either `addOrUpdateOngoingTx` or we leave it as is. It's a click away to understand what it is

noahjoeris (approved):
nice work @NunoAlexandre 
Left some small comments

app/src/components/SendButton.tsx
noahjoeris: But I was very confused to be honest. If we are talking about props in a component I am thinking of the props of THIS component. Not another one.

app/src/components/SendButton.tsx
NunoAlexandre: I forgot to commit & push - done now 👍

app/src/components/SendButton.tsx
NunoAlexandre: > Maybe also worth thinking about renaming the component from Button to BaseButton  or similar


I think that's unnecessary. `Button` is already a "base button", as the name suggests (in the sense that there is nothing specific about its name). `SendButton` is a clear subset of a `Button`. So I vote to keep it as is,

app/src/components/SendButton.tsx
NunoAlexandre: > Ahh I see. You are storing the props of the other button component.

Kinda. I am handling the props that make each button state unique or specific so that we can overwrite it and thus morph it.

app/src/components/Transfer.tsx
```diff
@@ -331,19 +332,21 @@ const Transfer: FC = () => {
           tokenAmount={tokenAmount}
           fees={fees}
           durationEstimate={durationEstimate}
+          className={transferStatus === 'Idle' ? '' : 'opacity-30'}
```
Victor-Salomon, line 335: (eventually to avoid passing the empty string)
import clsx from 'clsx';
className={clsx({ 'opacity-30': transferStatus !== 'Idle' })}

app/src/hooks/useParaspellApi.tsx
```diff
@@ -29,81 +30,113 @@ const useParaspellApi = () => {
...
+      ? 'Transfer a̶p̶p̶r̶o̶v̶e̶d rejected'
+      : 'Failed to submit the transfer'
+
+    captureException(e)
     addNotification({
-      message: 'Failed to submit the transfer',
+      message,
       severity: NotificationSeverity.Error,
     })
   }
```
Victor-Salomon, line 142: heavy to review but looks promising 🤩

app/src/store/ongoingTransfersStore.ts
```diff
@@ -48,19 +49,26 @@ export const useOngoingTransfersStore = create<State>()(
       transfers: [],
 
       // Actions
-      addTransfer: newOngoingTransfer => {
+      addOrUpdate: newOngoingTransfer => {
```
Victor-Salomon, line 52: Let's split this into two actions when we outsource the transfer storage from the user browser to another solution 🦾

Victor-Salomon (approved):
Rad one 💥
Not much to change! 
Can't wait to test it.

app/src/store/ongoingTransfersStore.ts
NunoAlexandre: yeah I agree. I can also split it now, the major motivation being performance but given that we shouldn't have many ongoing txs at once that should not be a practical problem

Victor-Salomon: Considering we will need to update later, I would keep it this way for now :)

app/src/hooks/useParaspellApi.tsx
NunoAlexandre: Yeah it's all cleaned up now, all error handling happening in one single place on this module

Victor-Salomon: I was speaking about the complete hook 🦾

app/src/components/SendButton.tsx
NunoAlexandre: I mean, it is the (overwritten) props of this component, but this component is a subset of another component. #allinterconnected

I renamed the function and added a docstring, let me know if that helps 👍

app/tailwind.config.ts
NunoAlexandre: yeah - used it for 1/2 animations on the `SendButton`. I tried using tailwind-animate instead but couldn't make it nearly as nice but open to try that to save on another dependency

NunoAlexandre:
> Rad one 💥 Not much to change! Can't wait to test it.

@Victor-Salomon I addressed the ticket to remove a failed ongoing tx [in this commit](https://github.com/velocitylabs-org/turtle/pull/194/commits/c3c37fec82d1e2e3dbbdc77341ab003a8b30c59f)

app/src/hooks/useParaspellApi.tsx
```diff
@@ -29,81 +30,125 @@ const useParaspellApi = () => {
...
+  ) => {
+    setStatus('Idle')
     console.log('Transfer error:', e)
+    const message = txWasCancelled(sender, e)
+      ? 'Transfer a̶p̶p̶r̶o̶v̶e̶d rejected'
+      : 'Failed to submit the transfer'
+
+    if (txId) {
+      removeOngoing(txId)
+    }
```
Victor-Salomon, line 147: 🦾

Victor-Salomon (approved):
All good to me!!💥

PR #195, opened 2024-11-26
fix: xcm ongoing progress bar and transfer status

Victor-Salomon:
This PR handles the XCM transfer progress. 
It also includes some code cleaning to a more generic & future-proof approach (vs the current Snowbridge-centric).


NunoAlexandre:
@Victor-Salomon do you want to mark this as "Ready for Review"?

app/src/hooks/useTransferProgress.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { StoredTransfer } from '@/models/transfer'
+import { Direction } from '@/services/transfer'
+import { EstimatedTransferDuration } from './useTransfersEstimate'
+
+const getTransferProgress = (
+  transferDate: Date,
+  transferDirection: Direction,
+  bridgeStatus: EstimatedTransferDuration,
+) => {
+  let bridgeTimestamp = 0
```
NunoAlexandre, line 11: 1. This is not a "timestamp", right? And we should include the units in the name and rename it for clarity

```suggestion
  let estimatedDurationMs = 0
```

2. Not sure if it's useful to have `bridgeStatus: EstimatedTransferDuration` as a parameter now since it will be a hardcoded value. So it would make more sense in my view to just reference it directly until we have plans to have a good API offering us this data dynamically.

3. With 2. in mind, I also think this `switch` logic should be moved to a separate function `getDurationEstimate(direction, estimatedDuration)` that would return the estimate in seconds. Then here this whole 13 lines of code becomes:

```
let estimatedDurationMs = getDurationEstimate(direction)
```

app/src/hooks/useTransferProgress.tsx
```diff
@@ -0,0 +1,80 @@
...
+      break
+    case Direction.WithinPolkadot:
+      bridgeTimestamp = bridgeStatus.xcmTransferStatus * 1000
+      break
+    default:
+      console.error('Unsupported direction to get transfer progress')
+  }
+
+  const transferTimestamp = new Date(transferDate).getTime()
+  const targetTimestamp = bridgeTimestamp + transferTimestamp
```
NunoAlexandre, line 27: It's easier to read this the way around:

```suggestion
  const targetTimestamp = transferTimestamp + estimatedDurationMs
```

app/src/hooks/useTransferProgress.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const currentTimestamp = new Date().getTime()
+
+  if (currentTimestamp > targetTimestamp) return 90
```
NunoAlexandre, line 30: What is this? 👀

app/src/hooks/useTransferProgress.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  // time already spent between transfer start & current time
+  const diffTimeSinceTransfer = currentTimestamp - transferTimestamp
```
NunoAlexandre, line 33: Just an idea, up to you :)

```suggestion
  const timePassed = currentTimestamp - transferTimestamp
```

app/src/hooks/useTransferProgress.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const progress = (diffTimeSinceTransfer / bridgeTimestamp) * 100
+
+  // To avoid displaying full progress bar, keep a 10% buffer.
+  // It returns 90% max and min 1% (to improve UI)
+  return Math.min(progress < 1 ? 1 : progress, 90)
```
NunoAlexandre, line 38: We can use the [`mod` function](https://stackoverflow.com/questions/17524673/understanding-the-modulus-operator) here to simplify this. And 1% as min is kinda too little, maybe ~5% if better?

app/src/hooks/useTransfersEstimate.tsx
```diff
@@ -0,0 +1,52 @@
+import { Direction } from '@/services/transfer'
+import useSnowbridgeContext from './useSnowbridgeContext'
+import { useQuery } from '@tanstack/react-query'
+import { getSnowBridgeEtimatedTransferDuration } from '@/context/snowbridge'
+import { Context } from '@snowbridge/api'
+
+export type EstimatedTransferDuration = {
+  ethBridgeStatus: number // to Ethereum
+  polkadotBridgeStatus: number // to Polkadot
+  xcmTransferStatus: number
```
NunoAlexandre, line 10: I left a comment above about renaming these fields

app/src/hooks/useTransfersEstimate.tsx
```diff
@@ -0,0 +1,52 @@
+import { Direction } from '@/services/transfer'
+import useSnowbridgeContext from './useSnowbridgeContext'
+import { useQuery } from '@tanstack/react-query'
+import { getSnowBridgeEtimatedTransferDuration } from '@/context/snowbridge'
+import { Context } from '@snowbridge/api'
+
+export type EstimatedTransferDuration = {
```
NunoAlexandre, line 7: I suggested above considering just making this a const.

```suggestion
export const ESTIMATED_DURATION = {
```

app/src/hooks/useTransfersEstimate.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const DEFAULT_CACHE_DURATION = 30 // minutes
+
+const DEFAULT_AVERAGE_TRANSFER_DURATION = {
+  ethBridgeStatus: 4 * 60 * 60, // 30 mins in seconds
+  polkadotBridgeStatus: 30 * 60, // 4h in seconds
+  xcmTransferStatus: 2 * 60, // 2 mins in seconds
+}
+
+const useTransferEstimate = (direction: Direction): EstimatedTransferDuration => {
```
NunoAlexandre, line 21: It's unclear what happens when this is called for XCM transfers 🤔

NunoAlexandre (commented):
Great clean up, Victor, and good seeing handled for XCM transfers. I left some comments but overall it's headed the right way 👌

NunoAlexandre:
> @NunoAlexandre, the PR was still a draft as I left a comment on Linear.
> 
> As we agreed on Linear: I will remove the useTransferEstimate hook that first fetches the estimation used to update the progress bar to use hardcoded values instead.
> 
> It loses a bit of precision, but it leans this heavy process.
> 
> I'll take into consideration all your naming suggestions and changes.

@Victor-Salomon perfect :) Merci bien 🙏

app/src/hooks/useTransferProgress.tsx
Victor-Salomon: This is when a transfer is taking more time than the estimated duration. 
When a transfer is blocked for any reason, usually 'an issue'.

app/src/hooks/useTransferProgress.tsx
Victor-Salomon: Hum, about modulo, can you explain how can it help here? 
(I know modulo, but I'm not sure to understand the need here).

Victor-Salomon: Yeah, return progress % 91 should work, _but it does not handle the minimum value._ 
I'll go for your second option 💥

PR #197, opened 2024-11-27
misc: Improve `formatAmount` & tweak ongoing and done dialogs

NunoAlexandre:
### Changes

- Improve `formatAmount` to not display too many significant numbers and offer a 'Shorter' and 'Longer' variants given that in some places we'd rather have a short display and in other places a longer one.
- Tweak the ongoing & done modals to use the TokenLogo instead of the token symbol
- Other small design tweaks and fixes

### Preview

**Before**
Screenshot 2024-11-27 at 11 12 14

Screenshot 2024-11-27 at 11 11 31

**After**
Screenshot 2024-11-27 at 11 15 40
Screenshot 2024-11-27 at 11 10 09


app/src/components/BetterImage.tsx
```diff
@@ -0,0 +1,23 @@
...
+  src: string
+  className?: string
+}
+
+/**
+ * Alternative to the default `Image` component that renders the image at the given `url` as a background image of a div,
+ * making it a lot easier to handle images with different ratios while forcing a fixed width and height.
+ *
+ */
+export const BetterImage: FC<BetterImageProps> = ({ width, height, src, className }) => {
```
Victor-Salomon, line 16: Can we rename it differently: 
TokenOverlayLogo, ImageBgOverlay or ImageOverlayWrapper?

Victor-Salomon (approved):
All good to me 🦾
I just left a naming suggestion.

app/src/components/BetterImage.tsx
NunoAlexandre: I love you very much but those names are awful. With that said, I was thinking of renaming this to just "Icon" or "Logo" since that's the only  purpose of this component

NunoAlexandre: Renamed it to Icon, thanks for chiming in 👌

PR #198, opened 2024-11-27
Add `ParaSpell` to the credits

NunoAlexandre:
Note: we definitely don't have to go with this rainbow tweak of mine - I just tried it out since that we now have the credits in two lines and with more text, which makes it heavier. By adding some color it makes it cooler and lighter but that's just my quick impression.

### Preview 

Screenshot 2024-11-27 at 14 47 11

### DIFF
--- a/app/src/app/globals.css
+++ b/app/src/app/globals.css
@@ -121,18 +121,7 @@
   }
 
   .rainbow-text {
-    background: linear-gradient(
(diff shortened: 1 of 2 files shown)


Victor-Salomon (approved):
I'm not quite sure about the rainbow text yet, but at least there are no code issues. 😅

PR #205, opened 2024-12-01
Improve fees loading

NunoAlexandre:
Improve the fees loading UI and UX:

- Add mini header "Loading fees" - otherwise users have no idea what they are waiting for
- Add a little emphatic and help tagline if it's taking too long

Eventually we want to redesign this but this feel - to me - like a small quick improvement for v2.

### Preview
(This demo was tweaked so that the fees kept loading if the amount == 5) 

https://github.com/user-attachments/assets/e7e9839c-dd33-4cac-b9c1-de1f2de767f4

### DIFF
--- a/app/src/components/Delayed.tsx
+++ b/app/src/components/Delayed.tsx
@@ -10,15 +10,14 @@ interface DelayProps {
 const Delayed: React.FC<DelayProps> = ({ millis, children }) => {
   const [shouldShow, setShouldShow] = useState<boolean>(false)
 
-  // Show when it reaches the delay timeout
+  // Should show when the delay timeout is reached
   useEffect(() => {
-    setTimeout(() => {
-      setShouldShow(true)
-    }, millis)
-  }, [millis, children])
+    const timer = setTimeout(() => setShouldShow(true), millis)
+    return () => clearTimeout(timer)
+  }, [millis])
 
   // Render
-  return shouldShow ? children : <></>
+  return shouldShow ? children : null
 }
 
 export default Delayed
--- a/app/src/components/Delayed.tsx
+++ b/app/src/components/Delayed.tsx
@@ -1,0 +1,24 @@
+import React, { useEffect, useState } from 'react'
+
+interface DelayProps {
+  /** The delay period in milliseconds */
+  millis: number
+  /** Children component(s) */
+  children?: React.ReactNode
+}
+
+const Delayed: React.FC<DelayProps> = ({ millis, children }) => {
+  const [shouldShow, setShouldShow] = useState<boolean>(false)
+
+  // Show when it reaches the delay timeout
+  useEffect(() => {
+    setTimeout(() => {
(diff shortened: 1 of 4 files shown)


app/src/components/Delayed.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+  // Show when it reaches the delay timeout
+  useEffect(() => {
+    setTimeout(() => {
+      setShouldShow(true)
+    }, millis)
+  }, [millis, children])
+
+  // Render
+  return shouldShow ? children : <></>
```
Victor-Salomon, line 21: I think `return shouldShow ? children : null` should work and is cleaner than returning an empty fragment.

NunoAlexandre: That didn't work, it resulted in a type error

app/src/components/Delayed.tsx
```diff
@@ -0,0 +1,24 @@
...
+  /** Children component(s) */
+  children?: React.ReactNode
+}
+
+const Delayed: React.FC<DelayProps> = ({ millis, children }) => {
+  const [shouldShow, setShouldShow] = useState<boolean>(false)
+
+  // Show when it reaches the delay timeout
+  useEffect(() => {
+    setTimeout(() => {
```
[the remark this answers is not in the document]
NunoAlexandre: For my own education, how is this a "clean up"? :)

app/src/components/Delayed.tsx
noahjoeris: thought the same

noahjoeris (approved):
Nice improvement 👏 
Victor left some comments

app/src/components/Delayed.tsx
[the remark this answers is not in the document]
Victor-Salomon: The cleanup function stops what the 'main' function is doing. It prevents perf issues when the component unmounts before the timeout completes, otherwise, it keeps it active.

app/src/components/Delayed.tsx
Victor-Salomon: Ok, weird but 👍

app/src/components/Delayed.tsx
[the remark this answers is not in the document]
NunoAlexandre: I see. I tested that and it doesn't seem like the timeout action takes effect if the component is unmount in the meantime. But I get the point, thanks for explaining, I will apply it 👍

Victor-Salomon (approved):
A nice UX improvement again 🦾

app/src/components/Delayed.tsx
NunoAlexandre: now it seems fine 🤷  maybe VSCode was acting up before or I was doing something stupid

NunoAlexandre:
> A nice UX improvement again 🦾

Thanks for the nice review @Victor-Salomon 🙏

PR #206, opened 2024-12-02
fix: tx summary fees format

Victor-Salomon:
Little commit/PR: I removed the NumberFlow animation from the TxSummary fees

### DIFF
--- a/app/src/components/TxSummary.tsx
+++ b/app/src/components/TxSummary.tsx
@@ -59,7 +59,7 @@ const TxSummary: FC<TxSummaryProps> = ({
                   </div>
                   {fees.inDollars > 0 && (
                     <div className="text-right text-turtle-level4">
-                      ${formatAmount(fees.inDollars, 'Long')}
+                      ${formatAmount(fees.inDollars, 'Short')}
                     </div>
                   )}
                 </div>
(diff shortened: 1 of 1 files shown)


app/src/components/TxSummary.tsx
```diff
@@ -60,7 +59,7 @@ const TxSummary: FC<TxSummaryProps> = ({
                   </div>
                   {fees.inDollars > 0 && (
                     <div className="text-right text-turtle-level4">
-                      <NumberFlow value={fees.inDollars} prefix="$" />
+                      ${formatAmount(fees.inDollars, 'Long')}
```
NunoAlexandre, line 62: I see that for the fees amount in the original token we don't pass `Long` so we shouldn't pass it here actually. Can you test a few scenarios and see how that looks? If it rounds it to `0` then we should of course keep it `Long` 👍

Victor-Salomon: I tried a few fees with either long or short. 
I have not seen the short option rounding to 0 => Let's go for this one.

NunoAlexandre: yeah it shouldn't because of the setting to have at least 2 significant digits 👍

NunoAlexandre (approved):

PR #207, opened 2024-12-02
Make Transfer button click-area bigger

rudolfs:
So that clicking anywhere on the button triggers navigation to the website.

I bumped into this a couple times and thought the button was broken. Also removed the `<button>` tag, because we can't nest a button within an `<a>` anchor tag if we want it to be valid HTML5.

Problem:

https://github.com/user-attachments/assets/d0e7c6eb-04e1-4d5a-83d8-0c6a83f5eb92

After the fix:

https://github.com/user-attachments/assets/46c65632-4008-4af4-8355-e7c859a0b8a8

### DIFF
--- a/site/src/components/header.tsx
+++ b/site/src/components/header.tsx
@@ -12,17 +12,17 @@ export default function Header() {
         </div>
       </Link>
 
-      <nav>
-        <ul className="items-center space-x-6">
-          <li>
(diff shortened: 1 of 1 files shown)


NunoAlexandre (approved):
Never noticed 🫨  thanks!!

noahjoeris (approved):

PR #209, opened 2024-12-03
fix: Track transactions only on production

rudolfs:
We don't want txs from devs testing locally to end up in the stats.
For reference: https://vercel.com/docs/projects/environment-variables/system-environment-variables#VERCEL_ENV.


NunoAlexandre (approved):
perfect 👌  thanks!!

noahjoeris (approved):

PR #210, opened 2024-12-03
fix: Show correct ethereum account identicon

noahjoeris:
This PR fixes an issue that evm addresses are just a purple circle and not the actual account logo (see below). And it fixes moonbeam's wrong `addressType`.

image

Before it was like this:
image


NunoAlexandre (approved):
LG!

PR #212, opened 2024-12-03
fix: Prevent fetching fees with an invalid manual address

NunoAlexandre:
Fixes https://linear.app/velocitylabs/issue/VEL-165/fix-error-when-trying-to-load-fees-with-an-invalid-dest-addr

When the dest address is set to manual and it's invalid, the app keeps trying to fetch fees the fees and it of course keeps failing and keeps piling up on error notifications:

Screenshot 2024-12-03 at 19 09 13

We also fix the design for the error state, making the input content reddish too.


Victor-Salomon (approved):
This looks all good to me!

PR #214, opened 2024-12-05
fix: scroll down on transfer initiation

noahjoeris:
This PR fixes a bug where it wasn't scrolled down to ongoing transfers correctly.


NunoAlexandre (approved):
Good catch! We don't need the delay like this but maybe it feels smoother UX wise 👍🏻

Victor-Salomon (approved):
LGTM! I guess the timeout allows first the ongoing transfers to exist :)

PR #219, opened 2024-12-09
Fix/filter nova wallet

Victor-Salomon:
Support displaying Nova Wallet from the app wallet by filtering through the browser window instead of Polkadot.
IMG_1547


NunoAlexandre (approved):
Nice 👌

PR #223, opened 2024-12-11
fix: remove not allowed tokens from list

Victor-Salomon:
Easy fix for this [issue](https://linear.app/velocitylabs/issue/VEL-194/filter-out-unsupported-tokens-instead-of-disabling-them)! 

=> Filter out unsupported tokens instead of disabling them


NunoAlexandre (commented):
This is fine but we should clean up the `allowed` field and type extension 🧹

Victor-Salomon:
> This is fine but we should clean up the `allowed` field and type extension 🧹

Hum, I'm not sure to understand this.

NunoAlexandre (approved):

noahjoeris (approved):

PR #225, opened 2024-12-11
feat: add moonbeam routes and refactor registry

noahjoeris:
This PR cleans up the registry by splitting it into multiple files. And I added Moonbeam routes
- USDT/C to AH
- DOT to RC
- GLMR to Hydra
- vDOT to Hydra works but tracking is broken. So I commented it out.


NunoAlexandre (commented):
Overall looking good but I am not convinced about the breakdown of the routes. I personally would prefer to keep them all in one place. Also if someone wants to look at this config to learn what routes we support there will be a single place to do that instead of several files.

The categorisation of "para to para" or "system to para" etc is not particularly useful since it doesn't solve any real problem; they are just transfers from A to B. The same goes for Snowbridge routes. So I'd rather keep them in a single place.

noahjoeris:
Alright got your point. The downsides I clearly saw with this approach are the following:
Bad overview, easy overseen merge errors (duplicates as we have seen before many times), correct route sorting breaks very fast and the order is faster messed up (as we have seen so far). 
A script to analyze them could definitely mitigate this.

noahjoeris:
removed the new files again

NunoAlexandre (approved):
🌟

PR #226, opened 2024-12-12
improvement ⚡ show precise available balance

noahjoeris:
This PR replaces our 95% quick fix for showing balances with the correct amount that is now fetchable via Paraspell.


NunoAlexandre (approved):
Voilá ✨

PR #227, opened 2024-12-12
improvement ⚡ Add snowbridge fee warning

noahjoeris:
This PR adds a warning that the snowbridge fees are only for its service and not eth gas costs.
Changed also the label Fee to Bridging Fee for such transfers.

image


NunoAlexandre (approved):
Nice 👌

PR #231, opened 2024-12-16
fix: walletconnect verify api

noahjoeris:
This PR tries to fix the reown cloud verify issue of "The source https://app.turtle.cool/ has not been authorized yet". I think the directory of the .well-known/walletconnect.txt was wrong, according to the docs [here](https://docs.reown.com/cloud/verify)

@Victor-Salomon Can you tell where you got the key from? I couldn't find it to verify if it's correct.


Victor-Salomon:
@noahjoeris, the last time I updated the well know key, it successfully verified the domain. 

I can't find anymore the key on Reown Cloud new interface, but I can see the domain is verified: 
image

You can use 1pw to connect [to the cloud here](https://cloud.reown.com/app/8b2e59d6-8b81-450f-9ce2-373bbabeccb6/project/c3d9cc5d-c641-42cd-99e0-bfc0df449e93)

Also, if you currently go the the app public file, you can read the key: https://app.turtle.cool/.well-known/walletconnect.txt

I do not know if it needs to be moved...

noahjoeris:
Hmm okay. Maybe it helped that I added the domain with a trailing `/` as seen in the error. But it's surprising that we get this error a lot on sentry.

Victor-Salomon:
Possibly the '/' could fix it... I do not know.
We can't click anymore on the "verify" button. 
Let's see if the error persists in the next days

noahjoeris:
Alright I will close this PR

PR #234, opened 2024-12-17
hotfix: temp disable some routes

noahjoeris:
Disables some moonbeam routes that we are currently fixing and make work.


Victor-Salomon (approved):
It looks good

PR #237, opened 2024-12-17
fix: moonbeam balances

Victor-Salomon:
Upgrade PS SDK and add moonbeam routes back. 
=> Balance fetches correctly now.


noahjoeris (approved):
Amazing. Good to have that back

PR #240, opened 2024-12-18
improvement: ocelloids tracking migration

Victor-Salomon:
This PR contains a migration to Ocelloids. 

Some paths are not supported by Ocelloids, including DOT transfers between parachains (non-system chains) and Interlay transfers. 
These unsupported paths are filtered out and tracked through our current Subscan implementation.

It also includes the outsourcing of the 'outdated transfers' in a dedicated hook (`useOngoingTransfersCleaner()`).

Finally, the PR updates the Progress bar design.


noahjoeris:
It's sooo smooth 😍

app/src/hooks/useOngoingTransferCleaner.tsx
```diff
@@ -0,0 +1,47 @@
...
+          tokenUSDValue: ongoing.tokenUSDValue ?? 0,
+          fees: ongoing.fees,
+          sender: ongoing.sender,
+          recipient: ongoing.recipient,
+          date: ongoing.date,
+          ...(explorerLink && { explorerLink }),
+        } satisfies CompletedTransfer)
+
+        addNotification({
+          message: 'Transfer verification failed.',
```
NunoAlexandre, line 36: The language here is not quite there. There is no "verification" taking place here, it's just tracking. And personally I prefer to not end these small and single lines with a `.`

```suggestion
          message: 'Transfer tracking failed',
```

NunoAlexandre:
@Victor-Salomon what's the status of this PR? what's keeping it from being ready for review? :)

Victor-Salomon:
@NunoAlexandre, I need to discuss it a bit with you, I also need to discuss some points with the team.

app/src/hooks/useParaspellApi.tsx
```diff
@@ -100,6 +100,7 @@ const useParaspellApi = () => {
                 ...(messageId && { parachainMessageId: messageId }),
                 ...(extrinsicIndex && { sourceChainExtrinsicIndex: extrinsicIndex }),
                 status: `Arriving at ${destinationChain.name}`,
+                onchainTimestamp: new Date(),
```
noahjoeris, line 103: Can we do that? To call it `onchain timestamp` but take the timestamp locally?

app/src/utils/ocelloids.ts
```diff
@@ -0,0 +1,240 @@
...
+                  xcm.XcmNotificationType.Timeout,
+                  remove,
+                  addCompletedTransfer,
+                  addNotification,
+                )
+                ws.close()
+                break
+              default: {
+                console.error('Unsupported XCM payload type')
+                break
```
noahjoeris, line 137: sentry log

noahjoeris (changes requested):
Awesome job @Victor-Salomon. That was heavy work. Thanks. Left some suggestions. 🚀

app/src/utils/ocelloids.ts
```diff
@@ -0,0 +1,240 @@
...
+import { Interlay } from '@/registry/mainnet/chains'
+import { Polkadot } from '@/registry/mainnet/tokens'
+
+type ResultData = {
+  message: string
+  severity: NotificationSeverity
+  status: TxStatus
+}
+
+const apiKey = process.env.NEXT_PUBLIC_OC_API_KEY_READ_WRITE || ''
```
NunoAlexandre, line 21: do we handle whether the keys is well loaded? We must report that case

app/src/utils/transfer.ts
```diff
@@ -295,3 +302,25 @@ export const startedTooLongAgo = (
...
+export const isParachain = (chain: Chain): boolean =>
+  chain.chainId !== AssetHub.chainId && chain.chainId !== RelayChain.chainId
+
+/**
+ * Checks if a both the source chain and the destination chain of a transfer are parachains
+ *
+ * @param transfer - The ongoing transfer to check.
+ * @returns A boolean indicating whether the transfers source and destination are parachains.
+ */
+export const parachainsOnly = (
```
NunoAlexandre, line 321:
```suggestion
export const betweenParachains = (
```

or `isParaToPara`

app/src/utils/ocelloids.ts
```diff
@@ -0,0 +1,240 @@
...
+      })
+
+    return OCLD_ClIENT.agent<xcm.XcmInputs>('xcm')
+  } catch (error) {
+    console.log(error)
+    captureException(error)
+  }
+}
+
+export const xcmOcceloidsSubscribe = async (
```
NunoAlexandre, line 79: Can you document this?

NunoAlexandre (commented):
Looks good overall. It would be nice to thrive for cleaner code but this is a solid start of this migration. I left a few comments, Noah's review as good already 👍

app/src/hooks/useOcelloidsSubscribe.tsx
```diff
@@ -0,0 +1,57 @@
...
+  xcmOcceloidsSubscribe,
+} from '@/utils/ocelloids'
+import { StoredTransfer } from '@/models/transfer'
+import useOngoingTransfers from '@/hooks/useOngoingTransfers'
+import useCompletedTransfers from '@/hooks/useCompletedTransfers'
+import useNotification from '@/hooks/useNotification'
+
+const useOcelloidsSubscribe = (ongoingTransfers: StoredTransfer[]) => {
+  const { remove } = useOngoingTransfers()
+  const { addCompletedTransfer } = useCompletedTransfers()
```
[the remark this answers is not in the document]
Victor-Salomon: Let's not handle those naming inconsistencies in this PR as it's not in the scope, and the PR is already heavy.

app/src/hooks/useParaspellApi.tsx
Victor-Salomon: I understand your comment. This key corresponds to the moment the transfer is finalized on the chain. (We do not have access to this data (date/timestamp) in the callback and it would be overkill to connect to the chain just for this). 

Any preference below?
  
- finalizedTimeStamp?
- finalizedAt?
- finalizedTxDate?
- finalizedDate?

app/src/utils/ocelloids.ts
Victor-Salomon: Now we do :)

app/src/utils/ocelloids.ts
Victor-Salomon: Done

app/src/hooks/useParaspellApi.tsx
NunoAlexandre: > finalizedAt?

my vote ✋

app/src/utils/transfer.ts
```diff
@@ -295,3 +302,25 @@ export const startedTooLongAgo = (
...
 }
+
+/**
+ * Checks if a chain is not a system chain (Asset Hub or Relaychain) but a parachain.
+ *
+ * @param chain - The chain to check.
+ * @returns A boolean indicating whether the chain is a system chain (Asset Hub or Relaychain) or a parachain.
+ */
+export const isParachain = (chain: Chain): boolean =>
+  chain.chainId !== AssetHub.chainId && chain.chainId !== RelayChain.chainId
```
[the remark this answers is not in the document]
noahjoeris: hmm I'm not sure. Maybe just add in the comments that it is incomplete.

NunoAlexandre (approved):
Nice :) a small step for humanity, a great step for Turtle 🐢

noahjoeris (approved):

app/src/utils/ocelloids.ts
```diff
@@ -102,36 +102,46 @@ export const xcmOcceloidsSubscribe = async (
...
-                //     addNotification,
-                //   )
-                //   ws.close()
-
-                // Do nothing?
+              case xcm.XcmNotificationType.Hop: {
+                const hopOutcome =
+                  payload.type === xcm.XcmNotificationType.Hop
+                    ? payload.waypoint.outcome
+                    : undefined
```
NunoAlexandre, line 114: @Victor-Salomon we already know from `case xcm.XcmNotificationType.Hop` that this is a `Hop`, so why check it here again?

Another wide comment is to not use ternary operators on enums. `XcmNotificationType` is not a boolean so using ternary operators against it might lead to silent bugs

NunoAlexandre (approved):
Looks good to me apart from the comment about the ternary operator :) Happy to merge this and work on that later 👍

app/src/utils/ocelloids.ts
Victor-Salomon: well spotted, I've updated it several times and forgot to remove it :)
Thanks!

app/src/utils/transfer.ts
```diff
@@ -295,3 +302,27 @@ export const startedTooLongAgo = (
       : thresholdInHours.bridge * 60 * 60 * 1000
   return new Date().getTime() - new Date(transfer.date).getTime() > timeBuffer
 }
+
+/**
+ * Checks if a chain is not a system chain (Asset Hub or Relaychain) but a parachain.
+ * This is a tmp helper used with in Ocelloids tracking to filter non supported paths.
+ * ⚠️ the helper does not verify every system chain.
```
noahjoeris, line 309: nice warning 😎

app/src/utils/ocelloids.ts
```diff
@@ -0,0 +1,286 @@
...
+      // Exclude transfer if DOT is transfered between parachains
+      if (isTransferringDotBetweenParachains(t)) return false
+      return true
+    }
+    return false
+  })
+}
+
+export const initOcelloidsClient = () => {
+  if (!OCELLOIDS_API_KEY || !OCELLOIDS_API_KEY.length)
```
noahjoeris, line 44:
```suggestion
  if (!OCELLOIDS_API_KEY)
```
an empty string is false

app/src/utils/ocelloids.ts
```diff
@@ -0,0 +1,286 @@
...
+                  ws.close()
+                }
+                break
+              }
+              case xcm.XcmNotificationType.Received: {
+                const finalOutcome =
+                  payload.type === xcm.XcmNotificationType.Received &&
+                  'outcome' in payload.destination
+                    ? payload.destination.outcome
+                    : undefined
```
noahjoeris, line 133: here as well

app/src/utils/ocelloids.ts
```diff
@@ -0,0 +1,286 @@
...
+    severity,
+    dismissible: true,
+  })
+
+  if (xcmMsgType === xcm.XcmNotificationType.Hop || xcmMsgType === xcm.XcmNotificationType.Timeout)
+    captureException(new Error(`Ocelloids tracking error:${message}`), {
+      tags: { XcmNotificationType: xcmMsgType, ...(transferOutcome && { transferOutcome }) },
+      extra: { transfer },
+    })
+}
```
noahjoeris, line 240: great

noahjoeris (approved):
amaaaaazing ❤️

app/src/utils/ocelloids.ts
Victor-Salomon: This is different: `outcome` needs to be narrowed down to `payload.destination`, which has 2 types possible.
If I remove the `undefined` return, the type will be `false | 'Success' | 'Fail'`.
Better like this since I handle `undefined` cases later.

PR #243, opened 2024-12-19
improvement: faster tx summary showing

noahjoeris:
This PR adds some tricks to show the txSummary faster. It uses fallback addresses to fetch fees if no addresses are inputed yet. After they are inputed it refetches the fees to be correct.

Closes VEL-197

### DIFF
--- a/app/src/utils/address.ts
+++ b/app/src/utils/address.ts
@@ -91,3 +91,13 @@ export const getRecipientAddress = (manualRecipient: ManualRecipient, wallet?: W
 /** Get the transfer sender address from the sender origin base (Substrate or Ethereum)*/
 export const getSenderAddress = async (sender: Sender): Promise<string> =>
   sender instanceof JsonRpcSigner ? await sender.getAddress() : (sender as InjectedAccount).address
+
+/** Get an example address for a specific type. Can be used to prefetch the fees before address input, as fees shouldn't differ. */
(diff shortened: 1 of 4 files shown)


Victor-Salomon (approved):
Smart tweak 😮‍💨

NunoAlexandre:
> After they are inputed it refetches the fees to be correct.

Fees don't change depending on the recipient address so I don't think that we need to refetch

app/src/utils/address.ts
```diff
@@ -91,3 +91,13 @@ export const getRecipientAddress = (manualRecipient: ManualRecipient, wallet?: W
 /** Get the transfer sender address from the sender origin base (Substrate or Ethereum)*/
 export const getSenderAddress = async (sender: Sender): Promise<string> =>
   sender instanceof JsonRpcSigner ? await sender.getAddress() : (sender as InjectedAccount).address
+
+/** Get an example address for a specific type. Can be used to prefetch the fees before address input, as fees shouldn't differ. */
+export const getExampleAddress = (type: AddressType): string => {
```
NunoAlexandre, line 96: The right word here is _placeholder_ 

```suggestion
/** Get a placeholder address for a specific type. Can be used to prefetch the fees before address input, as fees shouldn't differ. */
export const getPlaceholderAddress = (type: AddressType): string => {
```

NunoAlexandre (approved):
Sweet 🍯

PR #244, opened 2024-12-20
improvement: more robust tx analytics

rudolfs:
There were a couple issues with the current implementation:
  - we weren't guarding for the environment in all call sites -> moved the check inside the analytics module
  - we weren't tracking all the data to easily create badges of amount sums by token and show the token logo -> now we should have all the data needed to do aggregations per token, but also a total sum of all tokens
  - the document IDs in the collection was a mixture of tx IDs and random UUIDs which would make it harder to analyse -> now we just use firebase generated IDs for the documents and have a separate field for the tx ID if it's present
  - there was confusion on which ENV variables should be set for the firebase logging -> now we only have to set `NEXT_PUBLIC_FIREBASE_PROJECT_ID='app-turtle-cool'` on production, locally we don't need to set anything
  - also logging the URL the app is deployed on just in case someone else deploys it other than app.turtle.cool, so we can have more insights for debugging

Open questions:
  - should we actually log tx id, sender and receiver address or other tx identifying info or should we preserve anonymity?
  - do we care about showing token origin badge for the stats or is it enough to just show the token logo?
  - do we want to log fees?

Example transfer:
Screenshot 2024-12-20 at 15 49 46
What was logged:
Screenshot 2024-12-20 at 15 51 17


rudolfs:
Switched to draft because I haven't tested it yet (we don't have enough funds in the test wallet).

app/src/hooks/useParaspellApi.tsx
```diff
@@ -82,10 +81,12 @@ const useParaspellApi = () => {
             if (eventsData) {
               const { messageHash, messageId, extrinsicIndex } = eventsData
 
+              const id = getTxId(result)
```
NunoAlexandre, line 84:
```suggestion
              const id = getTxId(result)
```

NunoAlexandre (approved):
This is good 👌  are we able to fix the old entries on Firebase? I guess not since we didn't keep the `decimals`?

rudolfs:
> This is good 👌 are we able to fix the old entries on Firebase? I guess not since we didn't keep the `decimals`?

We probably could migrate the old entries with a script. Not sure it's worth the effort though. When I did the sum last time I wrongly used `usdValue`, so I assume the amounts are way less than I reported before. But I can look into it when I get back from vacay.

Victor-Salomon (approved):
LGTM 🦾

PR #247, opened 2025-01-06
Improvement/ps papi

Victor-Salomon:
updates Paraspell to the latest version and uses PAPI instead of PJS


app/src/utils/paraspell.ts
```diff
@@ -3,75 +3,52 @@ import { Token } from '@/models/token'
...
-  // to relay chain
-  else if (destinationChain.chainId === 0) {
-    return await Builder(api)
-      .from(sourceChainFromId as Exclude<TNodeDotKsmWithRelayChains, 'Polkadot' | 'Kusama'>)
-      .amount(amount)
-      .address(recipient)
-      .build()
-  }
-  // para to para
   else {
```
Victor-Salomon, line 33: the `else` statement is not useful enaymore no?

Victor-Salomon (commented):
Let's see the Interlay case before merging it. 
Otherwise, the PR seems good. 
Nice job @noahjoeris 🦾

app/src/utils/paraspell.ts
noahjoeris: yep. I removed it in the other branch :/

noahjoeris (approved):

PR #248, opened 2025-01-06
fix: fix isPreview and isProduction util functions

rudolfs:
When "Automatically expose System Environment Variables" is enabled in Vercel, all env variables have to be prefixed by `NEXT_PUBLIC_`. [Source](https://github.com/vercel/next.js/discussions/45802#discussioncomment-9022511).


NunoAlexandre (approved):
👌 nice catch

PR #250, opened 2025-01-08
feat: moonbeam traking with ocelloids

Victor-Salomon:
This PR supports tracking Moonbeam transfers using the EVM tx hash. The `getTxHashFromEvent()` helper sets the corresponding transaction hash while handling Moonbeam-specific exceptions.

I've also implemented a tweak/solution to update Moonbeam transfers status and start the progress bar accordingly.


app/src/utils/ocelloids.ts
```diff
@@ -277,3 +288,41 @@ const getNotification = (
...
+
+const getEvmTxHashFromEvent = (event: AnyJson): string | undefined => {
+  return event &&
+    typeof event === 'object' &&
+    'extrinsic' in event &&
+    event.extrinsic &&
+    typeof event.extrinsic === 'object' &&
+    'evmTxHash' in event.extrinsic
+    ? (event.extrinsic.evmTxHash as string)
+    : undefined
```
noahjoeris, line 308: Can't this be replaced with:
`return event?.extrinsic?.evmTxHash` as string | undefined

Victor-Salomon: Unfortunately no, I could not find an alternative to narrow it because of the AnyJson type provided by Ocelloids. 

See this error with `return event?.extrinsic?.evmTxHash`

image

noahjoeris (approved):

PR #256, opened 2025-01-17
fix: moonbeam xc token symbol issue

noahjoeris:
Fixes VEL-265

### DIFF
--- a/app/src/utils/paraspell.ts
+++ b/app/src/utils/paraspell.ts
@@ -47,10 +47,9 @@ export const getTokenSymbol = (sourceChain: TNodeDotKsmWithRelayChains, token: T
   let tokenSymbol: string | undefined
   if (sourceChain === 'Moonbeam') {
     tokenSymbol = supportedAssets.find(a => {
-      const strippedAsset = a.toLowerCase().startsWith('xc')
-        ? a.toLowerCase().slice(2)
-        : a.toLowerCase()
-      return strippedAsset === token.symbol.toLowerCase()
+      const lowered = a.toLowerCase()
+      const stripped = lowered.startsWith('xc') ? lowered.slice(2) : lowered
+      return stripped === token.symbol.toLowerCase()
     })
   } else tokenSymbol = supportedAssets.find(a => a.toLowerCase() === token.symbol.toLowerCase())
 
--- a/app/src/utils/paraspell.ts
+++ b/app/src/utils/paraspell.ts
@@ -44,7 +44,16 @@ export const createTx = async (
 export const getTokenSymbol = (sourceChain: TNodeDotKsmWithRelayChains, token: Token) => {
   const supportedAssets = getAllAssetsSymbols(sourceChain)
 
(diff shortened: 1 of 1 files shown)


NunoAlexandre (changes requested):
Appoligies if I am overseeing something here but I don't think this is the solution. We should use the `assetUid` map in the registry instead.

noahjoeris:
@NunoAlexandre I don't understand. Could you expand on your thought?

NunoAlexandre:
> @NunoAlexandre I don't understand. Could you expand on your thought?

Follow this hypothetical example:

Let's say the user wants to send polkadot native USDC from Moonbeam. So in this function you tweaked, we query Moonbeam's supported assets and look for a match by stripping the "xc" prefix. If there's a match, we will basically get `xcUSDC` (in this hypothetical example).

What I am suggesting is that we shouldn't do this at all. Instead, we should add a Moonbeam -> USDC -> xcUSDC map entry to the `assetUid` in the registry and make that O(1) query instead . That's exactly what that property of the registry is meant for.

Let me know if you need me to clarify or - again - if I am overseeing something here.

noahjoeris:
Ah okay I think I understand thanks. 
The thing is that Paraspell also just strips the 'xc' away at the moment 😅
So yeah I'll try to think about a solution. 
I don't see a strong argument for it at the moment tho but I agree it would be cleaner.

noahjoeris:
Just a side note: This whole function is only to spot differences in the Paraspell asset support and our registry asset support

noahjoeris:
Going to your mentioned O(1) is therefore only possible if we remove this check. Which makes us more error prone

NunoAlexandre:
Not sure if I follow but I trust you know what's best to do here now 👍

app/src/utils/paraspell.ts
```diff
@@ -44,7 +44,16 @@ export const createTx = async (
 export const getTokenSymbol = (sourceChain: TNodeDotKsmWithRelayChains, token: Token) => {
   const supportedAssets = getAllAssetsSymbols(sourceChain)
 
-  const tokenSymbol = supportedAssets.find(a => a.toLowerCase() === token.symbol.toLowerCase())
+  let tokenSymbol: string | undefined
+  if (sourceChain === 'Moonbeam') {
+    tokenSymbol = supportedAssets.find(a => {
+      const strippedAsset = a.toLowerCase().startsWith('xc')
+        ? a.toLowerCase().slice(2)
+        : a.toLowerCase()
```
NunoAlexandre, line 52: It would be nice to not repeat `toLowerCase` three times in a row 👍 

```suggestion
      const lowered = a.toLowerCase()
      const stripped = .startsWith('xc') ? lowered.slice(2) : lowered
     
```

NunoAlexandre (approved):
Small improvement but good otherwise 👍

noahjoeris:
Alright then let's keep it as I intended. If we see a need for using the xc prefix we now have a solution in mind. But so far there's none.

PR #257, opened 2025-01-22
feat: add xcm dry run api

noahjoeris:
This PR
- adds automatic dry run when available for xcm transfers

Design Decisions
- If no dry run is available we treat it as if it has succeeded
- The dry run returns precise xcm fees. But we run the dry run after the form was filled out and immediately before the tx-sign popup appears. So I decided to not use those fees for now. Alternatively, we can explore running the dry run as part of fetching the fees in txSummary. I would prefer to do that in a seperate PR tho.

### DIFF
--- a/app/src/hooks/useParaspellApi.tsx
+++ b/app/src/hooks/useParaspellApi.tsx
@@ -161,21 +161,19 @@ const useParaspellApi = () => {
     try {
       const result = await dryRun(params, params.sourceChain.rpcConnection)
 
-      const dryRunResult: DryRunResult = {
+      return {
         type: 'Supported',
         ...result,
       }
-
-      return dryRunResult
     } catch (e: unknown) {
       if (e instanceof Error && e.message.includes('DryRunApi is not available'))
-        return { type: 'Unsupported', success: false, failureReason: e.message } as DryRunResult
+        return { type: 'Unsupported', success: false, failureReason: e.message }
 
       return {
         type: 'Supported',
         success: false,
         failureReason: (e as Error).message,
-      } as DryRunResult
+      }
     }
   }
 
--- a/app/src/hooks/useParaspellApi.tsx
+++ b/app/src/hooks/useParaspellApi.tsx
@@ -4,15 +4,15 @@ import { getCachedTokenPrice } from '@/services/balance'
 import { SubstrateAccount } from '@/store/substrateWalletStore'
 import { getSenderAddress } from '@/utils/address'
 import { trackTransferMetrics } from '@/utils/analytics'
-import { createTx } from '@/utils/paraspell'
+import { handleObservableEvents } from '@/utils/papi'
+import { createTx, dryRun } from '@/utils/paraspell'
 import { txWasCancelled } from '@/utils/transfer'
 import { captureException } from '@sentry/nextjs'
+import { InvalidTxError, TxEvent } from 'polkadot-api'
 import { getPolkadotSignerFromPjs, SignPayload, SignRaw } from 'polkadot-api/pjs-signer'
 import useNotification from './useNotification'
 import useOngoingTransfers from './useOngoingTransfers'
 import { Sender, Status, TransferParams } from './useTransfer'
-import { InvalidTxError, TxEvent } from 'polkadot-api'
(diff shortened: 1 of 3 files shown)


NunoAlexandre:
@noahjoeris 

> If no dry run is available we treat it as if it has succeeded

this is not good. Using a `Result` type here would be best. If it ran, we check for the dry run result, if it failed we can for example capture the error with sentry and continue with the rest of the execution.

> The dry run returns precise xcm fees. But we run the dry run after the form was filled out and immediately before the tx-sign popup appears. So I decided to not use those fees for now. Alternatively, we can explore running the dry run as part of fetching the fees in txSummary. I would prefer to do that in a seperate PR tho.

Have you seen how different the xcm fees are? because I'd guess we are talking about a few cents difference at worse? dry running is probably a slow-enough operation to make this idea a no-go. Additionally, the user's wallet will also provide a fee estimate and that should be the last amount the user should consider. So I wouldn't spend time on this and run the risk of worsening the ux.

noahjoeris:
Could you look at the code? I implemented it how you described no?

NunoAlexandre:
> Could you look at the code? I implemented it how you described no?

yes but you used booleans and exceptions instead of a well typed `Result` approach. That's where my suggestion lays.

noahjoeris:
@NunoAlexandre Can you check again

app/src/hooks/useParaspellApi.tsx
```diff
@@ -155,22 +157,35 @@ const useParaspellApi = () => {
     }
   }
 
-  const validate = async (params: TransferParams) => {
+  interface DryRunResult {
+    dryRunSupported: boolean
+    success: boolean
+    error?: string
+    fee?: bigint
+  }
```
NunoAlexandre, line 165: This would be a lot nicer with a union type  👀 

```suggestion
  type DryRunResult = { type: 'Supported', result: boolean, fees: bigint } | { type: 'Unsupported', error?: string }
```

app/src/utils/paraspell.ts
```diff
@@ -79,6 +80,36 @@ export const moonbeamTransfer = async (
     .build()
```
Victor-Salomon, line 80: Not related, but can we remove the `console.log('Moonbeam transfer:', {...})` ?

app/src/hooks/useParaspellApi.tsx
Victor-Salomon: Another option could be to 'combine' PS `TDryRunResult` type instead of creating a new similar one? It might introduce more complexity to create the response though..
 
`type DryRunResult = { type: 'Supported' | 'Unsupported'} & TDryRunResult`

noahjoeris:
Thanks for your input guys. I updated it accordingly

app/src/hooks/useParaspellApi.tsx
```diff
@@ -152,6 +157,28 @@ const useParaspellApi = () => {
...
+  const validate = async (params: TransferParams): Promise<DryRunResult> => {
+    try {
+      const result = await dryRun(params, params.sourceChain.rpcConnection)
+
+      const dryRunResult: DryRunResult = {
+        type: 'Supported',
+        ...result,
+      }
+
+      return dryRunResult
```
NunoAlexandre, line 169: We can simplify this kind of stuff

```suggestion
      return {
        type: 'Supported',
        ...result,
      }
```

NunoAlexandre (approved):
Nice :) thanks for considering my feedback

app/src/hooks/useParaspellApi.tsx
noahjoeris: I overlooked this. Thanks

Victor-Salomon (approved):
Great work @noahjoeris 💥

NunoAlexandre (approved):

NunoAlexandre:
> ## Suspect Issues
> This pull request was deployed and Sentry observed the following issues:
> 
> * ‼️ **Error: Transfer dry run failed: Invalid checksum** `/` [View Issue](https://velocity-labs.sentry.io/issues/27202583/?referrer=github-pr-bot)
> 
> Did you find this useful? React with a 👍 or 👎

@noahjoeris  👀

PR #258, opened 2025-01-22
fix: ocelloids upgrade

Victor-Salomon:
It contains Ocelloids last version which fixes unsupported and missing tracking paths: 
- DOT btw Parachains
- PHA transfers

And includes: 
- Removing unused helpers
- Removing ongoing transfer filters

### DIFF
--- a/app/src/utils/ocelloids.ts
+++ b/app/src/utils/ocelloids.ts
@@ -21,16 +21,16 @@ enum xcmNotificationType {
   Bridge = 'xcm.bridge',
 }
 
-const OCELLOIDS_API_KEY = process.env.NEXT_PUBLIC_OC_API_KEY_READ_WRITE || ''
+export const OCELLOIDS_API_KEY = process.env.NEXT_PUBLIC_OC_API_KEY_READ_WRITE || ''
 
 // Helper to filter the subscribable transfers only
 export const getSubscribableTransfers = (transfers: StoredTransfer[]) =>
   transfers.filter(t => resolveDirection(t.sourceChain, t.destChain) === Direction.WithinPolkadot)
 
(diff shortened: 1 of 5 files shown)


noahjoeris (approved):
✨✨🧹🧹🧹

app/src/utils/ocelloids.ts
```diff
@@ -1,58 +1,44 @@
...
 
-const OCELLOIDS_API_KEY = process.env.NEXT_PUBLIC_OC_API_KEY_READ_WRITE || ''
-
-// TMP Helper until Ocelloids supports DOT transfers between non system chains.
-export const isTransferringDotBetweenParachains = (
-  transfer: StoredTransfer | OngoingTransferWithDirection,
-): boolean => {
-  return transfer.token.id === Polkadot.DOT.id && isParachainToParachain(transfer)
+enum xcmNotificationType {
+  Sent = 'xcm.sent',
```
NunoAlexandre, line 16: Question that I might answer myself later but that the introduction of this new enum interfere with old data on storage? I guess it doesn't matter as much since it's just about an ongoing tx.

NunoAlexandre (approved):
Nice 🧹

app/src/utils/ocelloids.ts
Victor-Salomon: Nope, the reason is the `xcmNotificationType` enum has been removed in Ocelloids SDK v2.4.x.  
I created this enum to avoid checking many times on the notification type since it is not typed anymore by Ocelloids.

Before v2.4.x: 
image

From v2.4.x: 
image

We never store this data, so no way to interfere with storage 🙂

PR #259, opened 2025-01-24
feat: reorder chains/tokens and add clear button

noahjoeris:
This PR
- reorders chains and tokens so the selected one is at the top in the dropdown
- enables clearing capability. I only enabled it for the dest chain as it only makes sense there.

Fixes VEL-224

image

### DIFF
--- a/app/src/components/Transfer.tsx
+++ b/app/src/components/Transfer.tsx
@@ -3,6 +3,7 @@ import useErc20Allowance from '@/hooks/useErc20Allowance'
 import useEthForWEthSwap from '@/hooks/useEthForWEthSwap'
 import useSnowbridgeContext from '@/hooks/useSnowbridgeContext'
 import useTransferForm from '@/hooks/useTransferForm'
+import { Chain } from '@/models/chain'
 import { resolveDirection } from '@/services/transfer'
 import { cn } from '@/utils/cn'
 import {
@@ -228,11 +229,11 @@ const Transfer: FC = () => {
               tokenAmount!.token,
             )
 
-            const reorderedOptions = reorderOptionsBySelectedItem(
-              options,
-              'uid',
-              destinationChain?.uid,
-            )
+            const reorderedOptions = reorderOptionsBySelectedItem<
+              Chain & {
+                allowed: boolean
+              }
+            >(options, 'uid', destinationChain?.uid)
 
             return (
               <ChainSelect
--- a/app/src/components/Transfer.tsx
+++ b/app/src/components/Transfer.tsx
@@ -218,6 +218,7 @@ const Transfer: FC = () => {
                   <WalletButton walletType={destinationChain?.walletType} />
                 )
               }
+              clearable
               walletAddress={destinationWallet?.sender?.address}
               className="z-30"
               disabled={transferStatus !== 'Idle' || !sourceChain || !tokenAmount?.token}
--- a/app/src/components/Transfer.tsx
+++ b/app/src/components/Transfer.tsx
@@ -10,6 +10,7 @@ import {
   getAllowedSourceChains,
   getAllowedTokens,
 } from '@/utils/routes'
+import { reorderOptionsBySelectedItem } from '@/utils/sort'
 import { formatAmount, getDurationEstimate } from '@/utils/transfer'
 import { Signer } from 'ethers'
 import { AnimatePresence, motion } from 'framer-motion'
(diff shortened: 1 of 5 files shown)


noahjoeris:
Let me know if you want the clear button to be designed differently.

NunoAlexandre:
> Let me know if you want the clear button to be designed differently.

It should be exactly how it's on Figma, with the only difference being that we don't include the 'X' 👍

app/src/components/Transfer.tsx
```diff
@@ -143,57 +144,74 @@ const Transfer: FC = () => {
...
-              placeholder="Source"
-              trailing={<WalletButton walletType={sourceChain?.walletType} />}
-              walletAddress={sourceWallet?.sender?.address}
-              className="z-50"
-              disabled={transferStatus !== 'Idle'}
-            />
-          )}
+          render={({ field }) => {
+            const options = getAllowedSourceChains(environment)
+            const reorderedOptions = reorderOptionsBySelectedItem(options, 'uid', sourceChain?.uid)
```
Victor-Salomon, line 149: Not a big deal, but reorderOptionsBySelectedItem expects a generic type no?

app/src/components/Transfer.tsx
```diff
@@ -143,57 +144,74 @@ const Transfer: FC = () => {
...
-                  }
-                />
-              }
-              className="z-40"
-            />
-          )}
+              }),
+            )
+
+            const reorderedOptions = reorderOptionsBySelectedItem(
```
Victor-Salomon, line 180: Same for the generic type comment in case you add it.

app/src/utils/sort.ts
```diff
@@ -0,0 +1,36 @@
...
+      if (value && typeof value === 'object') {
+        return (value as Record<string, unknown>)[key]
+      }
+      return undefined
+    }, obj) as string | undefined
+  }
+
+  const reorderedOptions: T[] = []
+  options.forEach(option => {
+    const uniqueId = getValue(option, key)
```
Victor-Salomon, line 27: getValue also expect a generic no?  `const uniqueId = getValue<T>(option, key)`

app/src/utils/sort.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (uniqueId === selectedId) {
+      reorderedOptions.unshift(option)
```
Victor-Salomon, line 29: 😎

Victor-Salomon (approved):
I've left some minor comments about the generics type. 
(I don't think our options have existing types, so feel free to ignore the one in the Transfer component). 

Otherwise, it looks nice and the selected tokens/chains as top value is mint 🤩

app/src/components/Transfer.tsx
noahjoeris: Typescript automatically detects the type

noahjoeris: Would look very ugly if I add it. I don't think it makes it any clearer.

```
const reorderedOptions = reorderOptionsBySelectedItem<
              Chain & {
                allowed: boolean
              }
>(options, 'uid', destinationChain?.uid)`

Comment (noahjoeris) [reply app/src/components/Transfer.tsx:149]: Could create a type like `ChainWithAllowedTag` and use that. But not sure if it's worth it only for this use case

Comment (Victor-Salomon) [reply app/src/components/Transfer.tsx:149]:
```
type Options = Chain & {
 allowed: boolean
}
const reorderedOptions = reorderOptionsBySelectedItem<Options>(options, 'uid', destinationChain?.uid)
```

This would be cleaner, but it's all good if it's naturally inferred ;)

app/src/utils/sort.ts
noahjoeris: not needed as well. But added it anyways :)

noahjoeris:
@NunoAlexandre Requesting approval from our leader

image

NunoAlexandre:
> @NunoAlexandre Requesting approval from our leader
> 
> image

@noahjoeris let's make it pixel-perfect as it is on Figma.

### Now
Screenshot 2025-01-27 at 12 08 55

### On Figma

- [ ] Let's add the 'X'
- [ ] Fix the padding
- [ ] Additionally, let's add a background change on hover (like we do for the wallet 'Connect' button for example)

Screenshot 2025-01-27 at 12 09 06

Thanks for hanging in there 🙏

noahjoeris:
Alright. Thought you told me without the X. Will add it.

NunoAlexandre:
> Alright. Thought you told me without the X. Will add it.

I did, I thought it would like nicer without it but I don't like it. Sorry, it's my bad Noah.

noahjoeris:
Updated El Patron @NunoAlexandre 

image

NunoAlexandre (approved):
♥️

PR #261, opened 2025-01-27
fix: selected token background

noahjoeris:
Colors the selected token background.


noahjoeris:
I skipped 2 nights to implement this

PR #263, opened 2025-01-31
improvement: registry unit tests

noahjoeris:
This PR
- removes circular dependencies
- removes token namespace. It’s an anti pattern and not recommended to be used anymore. And causes issues in tests.
- Adds static tests to our registry
- Adds Github Action to run tests as part of CI


app/src/__tests__/registry.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+describe('Routes', () => {
+  it('should have correctly structured routes', () => {
+    routes.forEach(route => {
+      expect(route).toHaveProperty('from')
+      expect(route).toHaveProperty('to')
+      expect(route).toHaveProperty('sdk')
+      expect(route).toHaveProperty('tokens')
+      expect(Array.isArray(route.tokens)).toBe(true)
+    })
```
NunoAlexandre, line 13: My strongly-typed heart 💔 🔪 

:)

app/src/__tests__/registry.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      tokenMap.forEach((assetUid, tokenId) => {
+        // Ensure the token ID (key) exists in the registry tokens
+        expect(Mainnet.REGISTRY.tokens.some(t => t.id === tokenId)).toBe(true)
+
+        // Ensure assetUid has either 'symbol' or 'id' property
+        expect('symbol' in assetUid || 'id' in assetUid).toBe(true)
+      })
+    })
+  })
+})
```
NunoAlexandre, line 93: very nice, @noahjoeris 🙏

app/src/registry/mainnet/tokens.ts
```diff
@@ -164,11 +164,11 @@ export namespace Eth {
...
       '{"parents":"2","interior":{"X2":[{"GlobalConsensus":{"Ethereum":{"chainId":"1"}}},{"AccountKey20":{"network":null,"key":"0x18084fbA666a33d37592fA2633fD49a74DD93a88"}}]}}',
     origin: snowbridgeWrapped(),
-  }
-}
+  },
+} as const satisfies Record<string, Token>
 
-export namespace Polkadot {
-  export const ACA: Token = {
+export const PolkadotTokens = {
```
NunoAlexandre, line 170: This seems unnecessary long, don't you think? Not a blocker of course but I liked the previous namespace layout better: e.g. `tokens.Polkadot.DOT`

app/src/registry/mainnet/tokens.ts
```diff
@@ -217,9 +217,9 @@ export namespace Polkadot {
       '{"parents":"1","interior":{"X2":[{"Parachain":"2031"},{"GeneralKey":"0x0001"}]}}',
     coingeckoId: 'centrifuge',
     origin: parachain(2031),
-  }
+  },
 
-  export const HDX: Token = {
+  HDX: {
```
NunoAlexandre, line 222: do I understand well that this means that these are no longer typed as `Token`?  if so, that's not so nice. I know we add tests to verify the presence of the expected fields but as bad as the type system in Typescript is, going this direction makes it even worse.

NunoAlexandre (commented):
Sweet refactor overall ✨ 

Left a question about the way this affects type-safety or not and a few small other remarks.

app/src/registry/mainnet/tokens.ts
noahjoeris: They should be even stronger typed now in fact.
Because of this: `} as const satisfies Record<string, Token>`

app/src/registry/mainnet/tokens.ts
noahjoeris: Hmm yes it's longer I know. But we should really avoid namespace. I need to think about alternatives.

app/src/__tests__/registry.test.ts
noahjoeris: 😆  Yeah tough to handle haha. It's just an insurance. Usually the type should show an error. BUT it is not a compiled language and therefore it runs also with errors. So it's security for when we overlook the type error, which already happened in the past!

NunoAlexandre: Definitely! With that, a remark: sometimes we add a field to types like `Token`. It would be cool if we could also add a test to verify that an old encoded `Token` can be decoded. We experienced issues when we released v2.0 and couldn't load old txs that had some missing fields that had been introduced in the meantime. We can do that in a follow up PR 👍

NunoAlexandre (approved):
🧹 ✨  🕺

app/src/__tests__/registry.test.ts
Victor-Salomon, line 93: Like them all as well 😎

app/src/registry/index.ts
```diff
@@ -47,56 +48,10 @@ export const SNOWBRIDGE_MAINNET_PARACHAIN_URLS = [
...
-  if (!token) throw Error(`Native Token for ${chain.uid} not found`)
-  return token
-}
-
-export function rpcConnectionAsHttps(rpc?: string): string {
-  if (!rpc) return ''
-  return rpc.replace('wss://', 'https://')
-}
-
 export function getAssetUid(
```
Victor-Salomon, line 51: Should we move this one to `registry/helpers.ts`?

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,40 @@
+name: Unit Tests
+
+on:
+  push:
```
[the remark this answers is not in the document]
noahjoeris: It runs on every commit on an opened PR. I think if we remove that it only runs on PR opening, and not again on updating the PR branch.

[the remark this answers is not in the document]
noahjoeris: It will basically behave the same way as our lint github action

app/src/registry/index.ts
noahjoeris: No because of circular dependencies

PR #264, opened 2025-02-03
improvement: add tx fees to ETH <> Polkadot env

Victor-Salomon:
This PR adds Ethereum tx fees when transferring ERC20 tokens from Eth to => Polkadot.

image

**New version:**
 
Capture d’écran 2025-02-04 à 20 06 50


noahjoeris:
Should we combine the fees into one UI element? Makes it cleaner. Like "Estimated fee: $1.34"

app/src/hooks/useFees.tsx
```diff
@@ -8,20 +8,26 @@ import { getCachedTokenPrice } from '@/services/balance'
...
+import { TokenAmount } from '@/models/select'
+import { ContractTransaction } from 'ethers'
 
 const useFees = (
   sourceChain?: Chain | null,
   destinationChain?: Chain | null,
-  token?: Token | null,
+  tokenData?: TokenAmount | null,
+  sourceWallet?: WalletInfo | undefined,
+  recipient?: string,
```
noahjoeris, line 27: Let's change `sourceWallet` to `senderAddress?: string`.

preferably for consistency also `recipient` to `recipientAddress`

noahjoeris: Could we also split up tokenData? We created the TokenAmount type for the purpose of controlling the react hook form. I would prefer to have such hooks be independent from that.
So I suggest those fields: `token?: Token` and `amount?: string`

noahjoeris (changes requested):
NIce feature mate 💯  Added some small recommendations.

NunoAlexandre:
@Victor-Salomon we shouldn't do this. The estimated fee should be the sum of the two sub-fees. In the future, we can design an expanded version of the `TXSummary` and display the full breakdown pieces but for now - as a Turtle design principle - we abstract that complexity away.

All users need to know here is how much it ill cost them in total to submit this transfer 👍

app/src/hooks/useFees.tsx
Victor-Salomon: Sure, I initially coded it like this and decided to update for one param. I just need to revert this [commit](https://github.com/velocitylabs-org/turtle/pull/264/commits/73ca1ce99e186e17968fa94f7ddc26ee85d2afef#diff-bbd36eb8d75ce195b8efeea2f61144b9855cc6ac8fb9feea7748eae6ac817852)

NunoAlexandre:
> The best option is to add/compute them in the TxSummary directly to keep flexibility later.

This 💯

> Should I keep it as it is on prod? We currently display "Bridge fees" and a Warning Tooltip.

The reason why we display "Bridging fees" for Snowbridge transfers is because until this latest version of their sdk we could only get the bridging fees and not the estimate for the whole transfer.

Now that we can, we should keep the same setup as we do for XCM transfers: Just show "fee" and show the total fee cost for the whole tx.

noahjoeris (approved):
Thanks for the changes 🚀

NunoAlexandre (commented):
Getting there :)

Still two things to change:

- "Bridging fee" => just "Fee"
- Drop the tooltip for bridging transfers since now we don't have to explain that it is just a bridging fee

PR #265, opened 2025-02-05
feat: add polimec routes

noahjoeris:
- add polimec chain and PLMC token
- add routes to/from AH with tokens DOT, USDT, USDC


NunoAlexandre (approved):
🚀

Victor-Salomon (approved):
Let's go 💥

PR #266, opened 2025-02-06
fix: unexpected fee triggers on source chain changes

Victor-Salomon:
Fixes the unexpected fee triggers on source chain changes. 

_**To reproduce the issue in production:**_
- Select a between parachains transfers ex: AH => USDC => Bifrost. 
- Update source chain to Eth. 
- The fees are triggered with the wrong data throwing an error. 

image

image


NunoAlexandre (approved):
Nice catch ✌️

noahjoeris (approved):
Nice

PR #268, opened 2025-02-10
fix: usdt revoke allowance

Victor-Salomon:
This PR fixes the USDT allowance issue. 

The USDT smart contract first requires the revoke of the current allowance before setting the new one. 
**It's a two-signature process.** 

I added the following extra info when needed: 
Capture d’écran 2025-02-10 à 10 43 09

### DIFF
--- a/app/src/hooks/useErc20Allowance.tsx
+++ b/app/src/hooks/useErc20Allowance.tsx
@@ -89,7 +89,6 @@ const useErc20Allowance = ({ network, tokenAmount, owner, context, refetchFees }
             await toPolkadot
               .approveTokenSpend(context, signer, tokenAmount!.token!.address, 0n)
               .then(x => x.wait())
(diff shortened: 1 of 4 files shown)


app/src/hooks/useErc20Allowance.tsx
```diff
@@ -78,6 +79,21 @@ const useErc20Allowance = ({ network, tokenAmount, owner, context, refetchFees }
       }
 
       try {
+        if (tokenAmount.token.id === EthereumTokens.USDT.id) {
+          if (!allowance) {
+            await fetchAllowance()
+          }
```
noahjoeris, line 85: That's better:
```suggestion
          if (allowance == null) await fetchAllowance()
```

noahjoeris (approved):
Well done 🏄

NunoAlexandre (approved):
This is good stuff, Victor! 👌

PR #269, opened 2025-02-10
chore: Update to Snowbridge v0.1.34

NunoAlexandre:


app/src/registry/index.ts
```diff
@@ -41,12 +41,20 @@ export const REGISTRY = {
...
 
-export const SNOWBRIDGE_MAINNET_PARACHAIN_URLS = [
-  rpcConnectionAsHttps(Mythos.rpcConnection),
-  rpcConnectionAsHttps(Bifrost.rpcConnection),
-  rpcConnectionAsHttps(Hydration.rpcConnection),
-  rpcConnectionAsHttps(Moonbeam.rpcConnection),
-]
+export const SNOWBRIDGE_MAINNET_PARACHAIN_URLS: { [paraId: string]: string } = {
+  // Asset Hub
+  '1000': rpcConnectionAsHttps(AssetHub.rpcConnection),
```
noahjoeris, line 46: Instead of hard coding all the para ids we could also read from the registry and dry.

noahjoeris (approved):
LGTM ♥️ left one small comment. Feel free to ignore it if you don't agree.

app/src/registry/index.ts
NunoAlexandre: I wanted to do that but unfortunately that's not possible because of the typing (that comes down from Snowbridge `[paraId: string]`). so it needs to be a hardcoded string. I can't reference the registry values 😢

Victor-Salomon: I think this should work: 

```
export const SNOWBRIDGE_MAINNET_PARACHAIN_URLS: { [paraId: string]: string } = {
  [AssetHub.chainId.toString()]: rpcConnectionAsHttps(AssetHub.rpcConnection),
  [BridgeHub.chainId.toString()]: rpcConnectionAsHttps(BridgeHub.rpcConnection),
  [Moonbeam.chainId.toString()]: rpcConnectionAsHttps(Moonbeam.rpcConnection),
  [Bifrost.chainId.toString()]: rpcConnectionAsHttps(Bifrost.rpcConnection),
  [Hydration.chainId.toString()]: rpcConnectionAsHttps(Hydration.rpcConnection),
  [Mythos.chainId.toString()]: rpcConnectionAsHttps(Mythos.rpcConnection),
};
```
or could be more dynamic like ? 

```
const SNOWBRIDGE_MAINNET_PARACHAINS = [AssetHub, BridgeHub, Moonbeam, Bifrost, Hydration, Mythos];
export const SNOWBRIDGE_MAINNET_PARACHAIN_URLS = Object.fromEntries(
  SNOWBRIDGE_MAINNET_PARACHAINS.map(chain => [chain.chainId.toString(), rpcConnectionAsHttps(chain.rpcConnection)])
);
```

Victor-Salomon: with relative imports obviously from our chains registry

Victor-Salomon:
@NunoAlexandre does this new 'context' include the Lazy init by default?

app/src/registry/index.ts
NunoAlexandre: @Victor-Salomon that's fine by me! can you apply those suggestions directly here so we can move this forward? Merci. 👍🏻

PR #270, opened 2025-02-11
fix: disabled send button

noahjoeris:
This PR fixes the issue that the send button is not enabled in some cases.


NunoAlexandre:
@noahjoeris can you explain the fix? The only meaningful change I see is the usage of `trigger` from `useForm`. Can you elaborate? It seems fine on my end but I want to understand

noahjoeris:
I figured out the issue only appears sometimes when the destination chain is updated. I simply added another form validation on destination chain change.

NunoAlexandre (approved):

PR #271, opened 2025-02-13
feat: One click Para -> Ethereum transfers

NunoAlexandre:
Support wETH from Hydration/BiFrost to Ethereum in one click  🚀


noahjoeris (approved):

app/src/app/api/bridging-fee/route.ts
```diff
@@ -0,0 +1,34 @@
+import { CACHE_REVALIDATE_IN_SECONDS } from '@/services/balance'
+import { getErrorMessage } from '@/utils/transferTracking'
+import { getParaEthTransferFees } from '@paraspell/sdk'
+import { unstable_cache } from 'next/cache'
+import { NextResponse } from 'next/server'
+
+/**
+ * Fetch the AH -> Ethereum bridging fee
```
Victor-Salomon, line 8: Fetch the _**Parachain**_ -> Ethereum bridging fee ?

app/src/utils/snowbridge.ts
```diff
@@ -241,7 +244,7 @@ export const getFeeEstimate = async (
           destinationChain.chainId,
           safeConvertAmount(amount, token) ?? 0n,
           bridgingFee.amount as bigint,
-          BigInt(0),
+          BigInt(3769142),
```
Victor-Salomon, line 247: Can we confirm here `destinationFeeInDOT` are this fixed amount `BigInt(3769142)`?

app/src/app/api/bridging-fee/route.ts
```diff
@@ -0,0 +1,34 @@
...
+ * It serves as a cached layer to reduce the latency of requesting this value.
+ *
+ * @returns the total amount the user will have to pay for execution and
+ * bridging fees on AH to send the tokens to Ethereum.
+ */
+export async function GET(_: Request) {
+  const fetchBridgingFee = unstable_cache(
+    async () => {
+      //todo(nuno): in the future we want to pass the granular fee breakdown (bridging + execution)
+      return (await getParaEthTransferFees()).reduce((acc, x) => acc + x)
```
Victor-Salomon, line 19: image

Local testings 👀

app/src/components/TxSummary.tsx
```diff
@@ -87,39 +86,65 @@ const TxSummary: FC<TxSummaryProps> = ({
...
+                    </div>
+                  )}
+                </div>
+                <div className="items-right flex">
+                  <div>
+                    <div className="flex items-center text-right text-xl text-turtle-foreground">
+                      {formatAmount(toHuman(additionalfees.amount, additionalfees.token))}{' '}
+                      {additionalfees.token.symbol}
+                    </div>
+
```
Victor-Salomon, line 131: UI issue here: 
image

app/src/utils/paraspell.ts
```diff
@@ -170,8 +177,19 @@ export function getNativeToken(chain: Chain): Token {
...
 
-export function getParaSpellNode(chain: Chain, relay: 'polkadot'): TNodeWithRelayChains | null {
+export function getParaSpellNode(chain: Chain): TNodeWithRelayChains | null {
   return chain.network === 'Ethereum' && chain.chainId === 1
     ? 'Ethereum'
-    : getTNode(chain.chainId, relay)
+    : getTNode(chain.chainId, 'polkadot')
+}
+
+/**
```
Victor-Salomon, line 186: Could we move this helper to `utils/address.ts` ?

app/src/hooks/useFees.tsx
```diff
@@ -29,8 +32,9 @@ const useFees = (
   recipientAddress?: string,
 ) => {
   const [fees, setFees] = useState<AmountInfo | null>(null)
-  const [ethereumTxfees, setEthereumTxFees] = useState<AmountInfo | null>(null)
+  const [bridgingFees, setBridgingFees] = useState<AmountInfo | null>(null)
```
Victor-Salomon, line 35: Just leaving a note since `ethereumTxFees` were => TxFees 
and `fees` were the => Bridging fees

So I feel this is confusing eventough it might be right 😅
image

NunoAlexandre: oh I see. Is this when sending from Ethereum -> AH/Parachain?

app/src/components/TxSummary.tsx
NunoAlexandre: And also an issue with the ETH amount 👀  👍

app/src/utils/snowbridge.ts
NunoAlexandre: No this is wrong, thanks for the catch

app/src/hooks/useFees.tsx
Victor-Salomon: Yes, it was wETH from ETH to AH (It must be the same path as the previous screenshot)

app/src/utils/transferTracking.ts
```diff
@@ -3,30 +3,37 @@ import { OngoingTransfers, StoredTransfer, TxTrackingResult } from '@/models/tra
...
-    transfers.push(...ethToParaTx)
+  const { toPolkadot, toEthereum, withinPolkadot } = ongoingTransfers
+
+  if (toPolkadot.length) {
+    for (const transfer of toPolkadot) {
+      const toParaTx = await history.toPolkadotTransferById(transfer.id) // must be {messageId_eq: "${id}", OR: {txHash_eq: "${id}"}
+      if (!toParaTx) continue
+      transfers.push(toParaTx)
+    }
   }
```
NunoAlexandre, line 22: We don't need the `if (toPolkadot.length) {`. `for (const transfer of toPolkadot) {` won't execute if `length` is 0

Victor-Salomon: You're right

app/src/app/api/bridging-fee/route.ts
NunoAlexandre: Yeah this returns the DOT fee that users will have to pay on AH to execute and bridge the tokens from AH to Ethereum 👍  Users need to have that amount of DOT on the source parachain which which will be sent to AH to cover those costs.

app/src/app/api/bridging-fee/route.ts
NunoAlexandre: @Victor-Salomon how did you execute this? `pnpm test` succeeds for me 👀

Victor-Salomon: @NunoAlexandre it's being displayed in the server/terminal console

NunoAlexandre: Thanks Victor!!!

app/src/app/api/bridging-fee/route.ts
noahjoeris, line 8: 8:18  Warning: tsdoc-escape-greater-than: The ">" character should be escaped using a backslash to avoid confusion with an HTML tag  tsdoc/syntax
```suggestion
 * Fetch the AH -\> Ethereum bridging fee
```

app/src/hooks/useTransferForm.tsx
```diff
@@ -215,6 +216,7 @@ const useTransferForm = () => {
         amount,
         recipient: recipient,
         fees,
+        bridgingFee: bridgingFee,
```
noahjoeris, line 219: 237:5  Warning: React Hook useCallback has a missing dependency: 'bridgingFee'. Either include it or remove the dependency array.  react-hooks/exhaustive-deps

app/src/app/api/bridging-fee/route.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Fetch the AH -> Ethereum bridging fee
+ *
+ * It serves as a cached layer to reduce the latency of requesting this value.
+ *
+ * @returns the total amount the user will have to pay for execution and
+ * bridging fees on AH to send the tokens to Ethereum.
+ */
+export async function GET(_: Request) {
+  const fetchBridgingFee = unstable_cache(
```
noahjoeris, line 16: Personal opinion: the name bridging fee is a very generic term imo that doesn't emphasize that it's an ethereum bridge, direction, or it uses snowbridge. Just a thought, no blocker

app/src/components/OngoingTransferDialog.tsx
```diff
@@ -175,6 +176,27 @@ export const OngoingTransferDialog = ({
...
+                    {formatAmount(
+                      toHuman(transfer.bridgingFee.amount, transfer.bridgingFee.token),
+                      'Long',
+                    )}{' '}
+                    {transfer.bridgingFee.token.symbol}
+                  </div>
+                  {typeof transfer.tokenUSDValue == 'number' && (
+                    <div className="text-turtle-level4">
+                      ${formatAmount(transfer.bridgingFee.inDollars, 'Long')}
+                    </div>
```
noahjoeris, line 195: If it's not of type number we don't show it? Can this happen or why do we have this check? Can't we just parse it to number in this case?

app/src/components/completed/TransactionDialog.tsx
```diff
@@ -175,6 +176,24 @@ export const TransactionDialog = ({ tx }: { tx: CompletedTransfer }) => {
...
+                <div className="font-bold">Bridging fee</div>
+                <div className="items-right flex flex-col space-x-1 text-right">
+                  <div className="text-lg">
+                    {formatAmount(toHuman(tx.bridgingFee.amount, tx.bridgingFee.token), 'Long')}{' '}
+                    {tx.bridgingFee.token.symbol}
+                  </div>
+                  {typeof tx.tokenUSDValue == 'number' && (
+                    <div className="text-turtle-level4">
+                      ${formatAmount(tx.bridgingFee.inDollars, 'Long')}
+                    </div>
```
noahjoeris, line 192: Same here

app/src/hooks/useFees.tsx
```diff
@@ -82,6 +85,30 @@ const useFees = (
           })
           setCanPayFees(info.sufficientForXCM)
 
+          // NOTE: when bridging from Parachain -> Ethereum, we have the local execution fees + the bridging fees.
+          // When bridging from AssetHub, the basic fees already take the bridging fees into account.
```
noahjoeris, line 89: That's such an important note that is hidden in the middle of the file. Maybe add doc comments to the state variables?

Victor-Salomon (approved):
Everything looks good to me :) 
LG!!! 🐢💥

app/src/store/completedTransfersStore.ts
```diff
@@ -7,6 +7,13 @@ interface CompletedTxState {
   addCompletedTransfer: (completedTransfer: CompletedTransfer) => void
 }
 
+const serializeFeeAmount = (fees: AmountInfo): AmountInfo => {
```
noahjoeris, line 10: nice

noahjoeris (commented):
Heavy work 💪  

I left some notes

app/src/hooks/useFees.tsx
NunoAlexandre: what do you mean specifically with "state variables"? 👀

app/src/components/OngoingTransferDialog.tsx
NunoAlexandre: also not sure why this was there in the first place. I deleted the 3 occurrences, if that causes trouble we will know 👍

app/src/app/api/bridging-fee/route.ts
NunoAlexandre: I don't quite agree with those exact arguments but I do think that it's too broad given that this is just to fetch the AH -> Ethereum bridging fee at the moment. In the future I want to pass arguments like source / dest chain so we can decide how and what to fetch.

app/src/hooks/useFees.tsx
noahjoeris: the useStates above

noahjoeris: Even better could be to define an interface for the useFees hook and add doc comments there to the bridging fee. This way using the hook shows the doc comment in the context menu.
Go as you like

NunoAlexandre: I moved this comment to the `Fee` type above. I think that's fine for now?

PR #273, opened 2025-02-17
fix: completed transfers storage

Victor-Salomon:
Converts Bigint to string to fix storing the completed transfers.


NunoAlexandre (approved):
✨

PR #274, opened 2025-02-18
Create Licenses folder and add Apache-2.0.txt

nicolares:
Need to add Apache license to Turtle for our DF. Can you please confirm I'm doing it in the right way? We also need to keep the MIT license and ideally we can add it under the LICENSES folder I created to keep it organized. But I don't know how to do that.

Thanks!


NunoAlexandre (approved):
@nicolares I applied the setup you suggested. Let me know if it looks good and nothing's missing.

PR #275, opened 2025-02-18
fix: useOngoingTransfersCleaner big int storage

Victor-Salomon:
Fix: Format fees amount to string to prevent bigint storage issues.


NunoAlexandre:
thanks @Victor-Salomon. I cherry picked the fix directly into my branch since I believe your based this branch off an old revision.

PR #279, opened 2025-02-20
feat ⚡ migrate to turbopack

noahjoeris:
This PR migrates turtle to use turbopack as the development bundler


Victor-Salomon (approved):
The code lgtm!
So if it works lets gooo 💥

PR #280, opened 2025-02-21
setup: add fonts & Hero UI

Victor-Salomon:
This PR adds Dazzed font & Hero UI lib.


noahjoeris (approved):
Good job. As discussed privately we could consider removing HeroUI and replace it with shadcn.

NunoAlexandre:
> Good job. As discussed privately we could consider removing HeroUI and replace it with shadcn.

do you guys want to write what pros and cons you both see in each?

noahjoeris:
The main reasoning is that we use HeroUI only for 2 components. And have the opportunity to remove a dependency that we dont have to maintain anymore.

PR #283, opened 2025-02-25
feat: xcm router sdk integration

noahjoeris:
- Adjusts hooks and models to our upcoming swaps feature
- Does not include UI changes


app/src/components/OngoingTransfer.tsx
```diff
@@ -35,7 +35,8 @@ const OngoingTransfer: FC<{
           color={colors['turtle-secondary']}
         />
         <p className="text-turtle-foreground)] no-letter-spacing text-xl font-normal">
-          {formatAmount(toHuman(transfer.amount, transfer.token))} {transfer.token.symbol}
+          {formatAmount(toHuman(transfer.amount, transfer.destinationToken))}{' '}
```
Victor-Salomon, line 38: this is temporary right? It will not work with the bridge and xcm no?

app/src/components/OngoingTransferDialog.tsx
```diff
@@ -80,8 +80,12 @@ export const OngoingTransferDialog = ({
               'xxl-letter-spacing flex items-center space-x-3 text-3xl leading-none text-turtle-secondary-dark sm:text-5xl'
             }
           >
-            <span>{formatAmount(toHuman(transfer.amount, transfer.token))}</span>
-            <TokenLogo token={transfer.token} sourceChain={transfer.sourceChain} size={40} />
+            <span>{formatAmount(toHuman(transfer.amount, transfer.destinationToken))}</span>
```
Victor-Salomon, line 83: the same with the current transfers we support that do not have a destination token?

app/src/hooks/useParaspellApi.tsx
```diff
@@ -109,75 +167,97 @@ const useParaspellApi = () => {
...
+
+    const tokenUSDValue = (await getCachedTokenPrice(params.sourceToken))?.usd ?? 0
+    const date = new Date()
+
+    const routerPlan = await createRouterPlan(params)
+    console.log('routerPlan: ', routerPlan)
+
+    const step1 = routerPlan.at(0)
+
+    step1?.tx
```
Victor-Salomon, line 190: is there a step2 somewhere? The code looks good, but it's hard to follow the path of what is happening 😅

Victor-Salomon: Maybe the step1 naming could be more explicit?

app/src/models/transfer.ts
```diff
@@ -15,7 +15,8 @@ export interface RawTransfer {
   destChain: Chain
   sender: string
   recipient: string
-  token: Token
+  sourceToken: Token
+  destinationToken: Token
```
Victor-Salomon, line 19: So all transfers now have a destinationToken field that can't be null | undefined ?

Victor-Salomon (approved):
Super solid! I do not see any 'wrong' things from a code point of view. 
It's a bit hard to catch the flow. 

One thing I might have not understood properly is how do we keep supporting the bridge and xcm transfers (mostly from a UI point of view) with the introduction of the destinationToken. 

Maybe we could schedule a call and you can show/explain the flow :) 

Nice job @noahjoeris 🧠💥

app/src/components/OngoingTransfer.tsx
noahjoeris: yep I included temporary UI changes. As I want to update the UI in different PRs

app/src/hooks/useParaspellApi.tsx
noahjoeris: No step2 yet. My first focus is to make it work with 1 step.

app/src/models/transfer.ts
noahjoeris: I was thinking about that yeah. Either we can make it nullable or use the same token as source and destination for usual transfers. What do you think @Victor-Salomon

noahjoeris: Same will be for sourceAmount and destinationAmount which I want to add in another PR. And maybe also sourceUsdValue and destinationUsdValue

Victor-Salomon: I would go with nullable fields @noahjoeris. 
In the end, this is what stick the most to the 'reality'. Bridge and XCM transfers do not have destinationToken while Swap transfers have some.

Unless it complicates this a lot. 
It might be a way to identify Swap transfer === when a transfer has both a source and a destination token.

PR #293, opened 2025-03-07
chore: Upgrade to Snowbridge Api v2

NunoAlexandre:


NunoAlexandre:
Still need to optimise the registry and the fees lookup but merging this now to hotfix ongoing bugs in production. On Monday I'll optimise it.

PR #295, opened 2025-03-10
imp: Move sb registry instantiation to `useSnowbridgeContext`

NunoAlexandre:
We get caching and unify both in one intersected type so that we don't have to keep track of two hooks and two values around the codebase. We might want to cache this server side in the future but for now caching client side for 12 hours should be fine.

### DIFF
--- a/app/src/hooks/useSnowbridgeContext.tsx
+++ b/app/src/hooks/useSnowbridgeContext.tsx
@@ -1,5 +1,7 @@
 import { getSnowBridgeContext } from '@/context/snowbridge'
 import useEnvironment from '@/hooks/useEnvironment'
+import { SnowbridgeContext } from '@/models/snowbridge'
+import { assetsV2 } from '@snowbridge/api'
 import { useQuery } from '@tanstack/react-query'
 
 const useSnowbridgeContext = () => {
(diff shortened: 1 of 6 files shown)


app/src/hooks/useSnowbridgeContext.tsx
```diff
@@ -12,9 +14,12 @@ const useSnowbridgeContext = () => {
...
     queryKey: ['snowbridgeContext', environment],
     queryFn: async () => {
-      return await getSnowBridgeContext(environment)
+      const ctx = (await getSnowBridgeContext(environment)) as SnowbridgeContext
+      ctx.registry = await assetsV2.buildRegistry(await assetsV2.fromContext(ctx))
+
+      return ctx
     },
-    staleTime: Infinity,
+    staleTime: 43200000, // 12 hours in milliseconds
```
Victor-Salomon, line 22: totally fine but would be even better to split it in a more readable way: 
`staleTime = 12 * 60 * 60 * 1000; // 12 hours in milliseconds`

app/src/registry/mainnet/routes.ts
```diff
@@ -62,7 +62,7 @@ export const routes: Route[] = [
   {
     from: AssetHub.uid,
     to: Ethereum.uid,
-    sdk: 'SnowbridgeApi',
+    sdk: 'ParaSpellApi',
```
NunoAlexandre, line 65: Note that I am moving away from the Snowbridge SDK for AH -> Ethereum and use ParaSpell instead. We already use ParaSpell for Para -> Ethereum and the goal was always to use PS for this too. I tried it and it worked just fine 👌

Victor-Salomon: Great! Thanks, @NunoAlexandre for the comment. 
I looked at the code to confirm tracking handles it! 
Even better if you tried it 😎

PR #297, opened 2025-03-11
feat: Add new UI for swap input

noahjoeris:
- creates new component ChainTokenSelect. Contains the new dropdown.
- removes old components: ChainSelect and TokenAmountSelect
- add priorityToken ordering. To make it easier to select the source token again
- update zod schema


app/src/components/ChainTokenSelect.tsx
```diff
@@ -0,0 +1,479 @@
...
+  const { data: ensAvatarUrl } = useEnsAvatar({
+    name: normalize(addressLookup || '') || undefined,
+  })
+
+  // Token hooks
+  const { price } = useTokenPrice(token.value)
+  const inDollars = !!amount?.value && price ? price * amount.value : undefined
+
+  // Search state
+  const [chainSearch, setChainSearch] = useState<string>('')
```
Victor-Salomon, line 100: Would, be great to switch chain and token search to a URL params base instead of the useState hook one. 
_For MVP this is totally good._

Victor-Salomon (approved):
Everything from a code perspective looks good. I can't wait to try it 🦾
I've left some comments with improvements that could be address later. 

Really nice job💥

app/src/components/ChainTokenSelect.tsx
noahjoeris: Hmm you think that makes sense? Imo it would make more sense for the selected chain/token but not for the filtered options.
Yeah this is post mvp thinking tho.

Victor-Salomon: Selecting a chain/token can be direct click or with search options no?

Victor-Salomon: I mean, in the end, either you select them by direct click or by searching in the searchbar no? where you could replace the search content with the token name for example?

noahjoeris: Nope. The search bar is just to filter the options. Not to select right now.

PR #298, opened 2025-03-12
fix: Validate SB-based transfers before estimating execution fees

NunoAlexandre:
Fixes https://velocity-labs.sentry.io/issues/29823189/replays/

Basically validate the transfer before going ahead with estimating execution fees which fails if the tx is invalid.

### DIFF
--- a/app/src/utils/snowbridge.ts
+++ b/app/src/utils/snowbridge.ts
@@ -119,8 +119,8 @@ export const getFeeEstimate = async (
           },
           tx,
         )
-        
-        if (checkValidation(validation) === 'Failed')
+
+        if (findValidationError(validation))
           return {
             origin: 'Ethereum',
             bridging: bridgingFee,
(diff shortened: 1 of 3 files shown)


app/src/utils/snowbridge.ts
```diff
@@ -133,3 +154,9 @@ export const getFeeEstimate = async (
       throw new Error('Unsupported direction for the SnowbridgeAPI')
   }
 }
+
+export const findValidationError = (
+  validation: toPolkadotV2.ValidationResult | toEthereumV2.ValidationResult,
+): toPolkadotV2.ValidationLog | toEthereumV2.ValidationLog | undefined => {
+  return validation.logs.find(log => log.kind == toPolkadotV2.ValidationKind.Error)
+}
```
noahjoeris, line 162: Does that also work for `toEthereumV2.ValidationResult` as a paremeter input?
Because we have this Polkadot specific check: `validation.logs.find(log => log.kind == toPolkadotV2.ValidationKind.Error`

NunoAlexandre: Good catch!!!

NunoAlexandre: Actually, the type system complains if we had that extra check because the two `ValidationError` types are identical 👇  given that we are dropping `toEthereum` and only use ParaSpell for that, this is fine as is 👍 

Screenshot 2025-03-12 at 21 34 16

PR #300, opened 2025-03-14
Make `ChainSelect` address copyable

NunoAlexandre:
### Changes

- Make `ChainSelect` address copyable
- Fix dropdown clickable area to only the chain and the `ChevronDown`

### DIFF
--- a/app/src/components/ChainSelect.tsx
+++ b/app/src/components/ChainSelect.tsx
@@ -18,6 +18,7 @@ import ChevronDown from './svg/ChevronDown'
 import { Cross } from './svg/Cross'
 import { Tooltip } from './Tooltip'
 import VerticalDivider from './VerticalDivider'
+import CopyAddress from './ClipboardCopy'
 
 interface ChainSelectProps extends SelectProps<Chain> {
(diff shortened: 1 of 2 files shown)


app/src/components/ChainSelect.tsx
```diff
@@ -92,36 +93,37 @@ const ChainSelect = forwardRef<HTMLDivElement, ChainSelectProps>(
...
-                <>
-                  <Image
-                    src={value.logoURI}
-                    alt={value.name}
-                    width={24}
-                    height={24}
-                    className="h-[2rem] w-[2rem] rounded-full border-1 border-turtle-foreground bg-background"
-                  />
+            <div className="flex h-[3.5rem] flex-grow items-center">
+              <div onClick={handleTriggerClick} className="flex flex-row items-center">
```
NunoAlexandre, line 98: The change here boils down to wrapping the chain and the ChevronDown components within a common `div` and have the `onClick={handleTriggerClick} ` on those two elements only. This prevents the dropdown from opening when the user misclicks the Connect/Disconnect button as well as copying the address now.

PR #302, opened 2025-03-14
Support ETH bridging

NunoAlexandre:


NunoAlexandre:
@noahjoeris @Victor-Salomon can you guys review this very soon? By supporting ETH, we need to take in consideration when sending from Ethereum, the transfer token and the fee token are the same so we need. An issue with the current setup is that `useFees` depends on `useBalance` while `useBalance` would also need to know the fees in order to return a balance amount that takes in consideration the amount needed to cover the fees. For ETH now I apply a 90% factor on the base balance which works fine now but we need to find a more resilient approach.

app/src/hooks/useBalance.ts
```diff
@@ -55,7 +55,8 @@ const useBalance = ({ env, chain, token, address }: UseBalanceParams) => {
               : (await fetchErc20Balance()).data
 
           if (fetchedBalance)
-            fetchedBalance.formatted = toHuman(fetchedBalance.value, token).toString() // override formatted value
+            // apply a 90% factor to safe-guard for fees and other unseen costs
+            fetchedBalance.formatted = (toHuman(fetchedBalance.value, token) * 0.9).toString() // override formatted value
```
noahjoeris, line 59: In my opinion this is mixing responsiblities. useBalance should just return the actual balance. It has nothing to do with fees. 
There is probably a better way.
If you want to keep it for now, please add a TODO and add a Linear ticket.

noahjoeris (approved):
Left one comment. 
Looks good otherwise 🙏 🔥

Victor-Salomon (approved):
Looks good to me aswell except the useBalance that should not stay like this too long :)

app/src/hooks/useBalance.ts
NunoAlexandre: I agree with you @noahjoeris 👍  ticket created: https://linear.app/velocitylabs/issue/VEL-365/decouple-usebalance-and-usefees

PR #305, opened 2025-03-19
improvement: update registry with multilocations

noahjoeris:
It adds multilocation objects to the registry and one helper function to get a token with a multilocation


noahjoeris:
I noticed vETH is not used at all. Should we remove it? @NunoAlexandre @Victor-Salomon

app/src/models/schemas.ts
```diff
@@ -16,7 +17,7 @@ export const tokenSchema: z.ZodType<Token> = z.object({
   symbol: z.string(),
   decimals: z.number(),
   address: z.string(),
-  multilocation: z.string(),
+  multilocation: z.custom<TMultiLocation>(),
```
NunoAlexandre, line 20: Won't this cause issues with backwards-compatibility?

app/src/models/token.ts
```diff
@@ -27,3 +32,16 @@ export type Origin = { type: 'Ethereum'; bridge: Bridge } | { type: 'Polkadot';
 export function getCoingekoId(token: Token): string {
   return token.coingeckoId ?? token.name.toLocaleLowerCase().replaceAll(' ', '-')
 }
+
+export function getTokenByMultilocation(multilocation: TMultiLocation): Token | undefined {
+  const token = REGISTRY.tokens.find(token => deepEqual(token.multilocation, multilocation))
+
+  // Usually a token should be found, so log a warning if not
+  if (!token)
+    captureException(new Error('Token multilocation not found'), {
```
NunoAlexandre, line 41:
```suggestion
    captureException(new Error('Token not found by multilocation'), {
```

NunoAlexandre (approved):
The changes look good :)

app/src/models/schemas.ts
noahjoeris: hm I am not sure honestly. We never used the multilocation. I'll verify

Victor-Salomon: It should be good. 
The final zod `schema` is used to resolve the Form Data. Nothing should be stored with this Schema type.

app/src/models/token.ts
```diff
@@ -7,7 +12,7 @@ export interface Token {
   address: string
   // The xcm multilocation of an asset serves as a contextual identifier of said asset.
   // Learn more at https://guide.kusama.network/docs/learn/xcm/fundamentals/multilocation-summary
-  multilocation: string
+  multilocation: TMultiLocation
```
Victor-Salomon, line 15: here, it might be worth checking backward compatibility since we registered all the completed transfers with the Token type where the multilocation is a string.

app/src/models/token.ts
```diff
@@ -27,3 +32,16 @@ export type Origin = { type: 'Ethereum'; bridge: Bridge } | { type: 'Polkadot';
 export function getCoingekoId(token: Token): string {
   return token.coingeckoId ?? token.name.toLocaleLowerCase().replaceAll(' ', '-')
 }
+
+export function getTokenByMultilocation(multilocation: TMultiLocation): Token | undefined {
```
Victor-Salomon, line 36: could we move this function out of the models folder/files? 
(same for the getCoingekoId())
Either we can create a `tokens.ts` file in utils or a more generic `helpers.ts`?

Victor-Salomon (approved):
LGTM !! 🦾

app/src/models/schemas.ts
noahjoeris: yep I tested it. All good

app/src/models/token.ts
noahjoeris: It works as we haven't used the multilocation yet

app/src/models/token.ts
noahjoeris: ✅

PR #306, opened 2025-03-19
Remove inline object literals from components props

jonidelv:
This PR addresses a potential performance issue by removing inline object literals from component props.

## Problem
Inlining object literals as props (like `<Component data={{key: value}}/>`) creates a new object instance on every render because JavaScript compares objects by reference, not value. This causes:

1. Unnecessary re-renders: Components using React.memo or PureComponent will re-render even when the actual data hasn't changed.
2. Increased memory usage: Each render creates new object instances that must be garbage collected.
3. Reduced performance: Additional rendering cycles and garbage collection affect application speed.

The same issue occurs when inlining functions `<Button onClick={() => handleClick()}/>` or other components as props `<TabPanelheader={<CustomHeader title="Dashboard" />} />` but these could be addressed in a separate PR.

## Solution
This PR moves inline object literals to component-level constants or file variables:

* Before: `<UserCard settings={{theme: 'dark', showAvatar: true}} />`
* After (outside component): 
  ```jsx
  // Constants defined outside component maintain reference across renders
  const staticSettings = {theme: 'dark', showAvatar: true};
  
  function MyComponent() {
    return <UserCard settings={staticSettings} />;
  }
  ```
* After (with useMemo - for dynamic values):
  ```jsx
  function MyComponent({ userId, isDarkMode }) {
    // Use useMemo when object depends on props or state
    const userSettings = useMemo(() => ({
      theme: isDarkMode ? 'dark' : 'light',
      userId: userId,
      showAvatar: true
    }), [isDarkMode, userId]);
    
    return <UserCard settings={userSettings} />;
  }
  ```

This ensures stable object references between renders, preventing unnecessary re-renders and improving memory efficiency.

### DIFF
--- a/app/src/components/WalletButton.tsx
+++ b/app/src/components/WalletButton.tsx
@@ -15,6 +15,12 @@ interface WalletButtonProps {
   className?: string
 }
 
+const walletButtonAnimationProps = {
+  initial: { opacity: 0 },
+  animate: { opacity: 1 },
+  exit: { opacity: 0 },
+}
+
 /** Wallet button component that is intended to support connecting to various different networks based on its address type. */
 const WalletButton = ({ walletType, className }: WalletButtonProps) => {
   const {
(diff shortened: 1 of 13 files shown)


app/src/components/WalletButton.tsx
```diff
@@ -15,6 +15,12 @@ interface WalletButtonProps {
   className?: string
 }
 
+const walletButtonAnimationProps = {
```
NunoAlexandre, line 18: how do you feel about reducing the naming redundancy on these? this would just be `animationProps`, given that "walletButton" is redundant as it's already within the `WalletButton` component. This would help with readability imo. what do you think?

NunoAlexandre (approved):
I like this a lot :) including the write up ✨

PR #308, opened 2025-03-21
Refactor FC usage, clean up component, and code styling

jonidelv:
Standardized `FC` usage across components for better consistency. Simplified and unified props definitions by introducing interfaces where applicable. Cleaned up redundant or unnecessary JSX syntax.

### DIFF
--- a/app/src/app/error.tsx
+++ b/app/src/app/error.tsx
@@ -1,12 +1,16 @@
 'use client'
-import { useEffect } from 'react'
+import { FC, useEffect } from 'react'
 import Image from 'next/image'
 import Link from 'next/link'
 import { captureException } from '@sentry/nextjs'
 
 const imageStyles = { objectFit: 'cover' as const }
 
-const Error = ({ error }: { error: Error & { digest?: string } }) => {
+interface ErrorProps {
+  error: Error & { digest?: string }
+}
+
+const Error: FC<ErrorProps> = ({ error }) => {
   useEffect(() => {
     captureException(error)
   }, [error])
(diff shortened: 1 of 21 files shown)


app/src/app/error.tsx
```diff
@@ -1,12 +1,16 @@
...
 import Link from 'next/link'
 import { captureException } from '@sentry/nextjs'
 
 const imageStyles = { objectFit: 'cover' as const }
 
-const Error = ({ error }: { error: Error & { digest?: string } }) => {
+interface ErrorProps {
+  error: Error & { digest?: string }
+}
+
```
jonidelv, line 12: I feel this was more readable, feel free to discard it if you think it is not

app/src/app/page.tsx
```diff
@@ -1,9 +1,7 @@
-import { FC } from 'react'
-
 import { HomeComponentSelect } from '@/components/HomeComponentSelect'
 import { TurtlesBackground } from '@/components/TurtlesBackground'
 
-const App: FC = async () => {
+const App = () => {
```
jonidelv, line 4: No need to use async here since we’re not awaiting any promises, and the child components are client components.

app/src/components/Account.tsx
```diff
@@ -5,21 +5,23 @@ import { cn } from '@/utils/cn'
...
   size = 14,
-}: {
-  network: Network
-  addressType?: AddressType
-  address: string
-  className?: string
-  allowCopy?: boolean
-  size?: number
-}) {
+}: AccountProps) {
```
jonidelv, line 24: I feel that when there are more than two props or if the prop types are too long, it’s more readable to define them using a type or an interface. Open to discussion.

noahjoeris (approved):
Lots of cleanups 🧹🧹🧽🧽🫧🫧
Good job 🔥

PR #314, opened 2025-03-26
feat: add swap output amount

noahjoeris:
This PR 
- Implements a function and hook to retrieve the swap output amount. Adds caching
- fixes some route issues
- fixes lint issues


app/src/hooks/useSwapOutputAmount.tsx
jonidelv, line 1: This file currently uses a `.tsx` extension even though it doesn’t include any JSX (UI/React HTML) syntax. Would it make sense to change it to `.ts` to better reflect its content?

jonidelv (approved):
Just a minor comment regarding a file extension, but LGTM

app/src/hooks/useSwapOutputAmount.tsx
noahjoeris: We could theoretically. Althought I like the consistency among the hooks.

PR #316, opened 2025-03-28
Switch from Dialog to Dynamic try comp. in Transaction History

jonidelv:
### Overview

This PR replaces the transaction detail modal with a dynamic tray component (`DynamicTray.tsx`) that swaps two UIs with an animation, resulting in a more compact UX ideal for the widget.

### Changes

- **Removed:** `TransactionDialog.tsx`.
- **Added:** Generic `DynamicTray.tsx` component.
- **Integrated:** Dynamic tray into Transaction History.
- **Adjusted:** Transaction History detail UI to be more compact and fit within the app container.


NunoAlexandre:
@jonidelv awesome to see this gaining shape :) 

I know that you don't have a useful support design to build this against and that's my bad. Let's work it through. My idea with this is not to have the expanded view consuming the entire Transfer form component's real estate. The idea was that the selected ongoing/done tx would expand in, almost like a scale up with more data that then the user could click outside of to close. So no back buttons, no taking the full real estate. Just an expanded/scaled up type of morphing experience.

This is a totally crappy sample of what I mean but it will help visualise it:

Screenshot 2025-03-28 at 11 25 05


So, ideally the expanded view would take 2-3x the height of the compact view. We need to redesign those extended/modals because now they look pretty cramped and ugly and won't work for this purpose.

PR #318, opened 2025-03-30
impr: make currencyTo use multilocation as well

noahjoeris:
The currencyTo needed special treatment before as it couldn't use parapsells token id. This PR makes the destination token use the multilocation or symbol now as well.


NunoAlexandre (approved):
🧹

PR #320, opened 2025-04-02
Refactor Dialogs/Modals for a more Compact Interface

jonidelv:
Updated OngoingTransferDialog and TransactionDialog to deliver a more compact UI that occupies less space and adapts better for animations transitioning from cards.


noahjoeris (commented):
LGTM. Just change the block explorer button again. 

And I recently realized it's pretty nice to break such large components down into subcomponents for better readability. In the same file. So you have a main component that is pretty short containing these subcomponents. What do you think?

PR #321, opened 2025-04-02
Rename hook and config files to use .ts extension

jonidelv:
Updated hook and config file extensions from .tsx to .ts to indicate that these files focus on logic rather than UI. This way, you can quickly spot which files contain JSX, and your editor even shows the right icons, making it easier to differentiate between UI and non-UI files at a glance.


Victor-Salomon (approved):
No change, just extension update! All good to me ✅

noahjoeris (approved):

PR #322, opened 2025-04-02
fix: remove issues on selecting chains and tokens

noahjoeris:
- fixes switch button
- fixes unallowed route issues
- upgrades paraspells to fix swap issues
- refactors some code


Victor-Salomon (approved):
Clean one 🦾

PR #325, opened 2025-04-03
impr: update transfer model for swap

noahjoeris:
- Adds source-/destinationTokenUSDValue
- Adds source-/destinationAmount


app/src/hooks/useOngoingTransfersTracker.ts
```diff
@@ -155,7 +158,36 @@ const useOngoingTransfersTracker = (ongoingTransfers: StoredTransfer[]) => {
         }
       } else {
         // ongoing transfer not found. This means it is more than 2 weeks old.
-        // TODO: handle this case
+        // Clean up the transfer and mark it as undefined
```
Victor-Salomon, line 161: I'm not sure about this. 

There is already the transferCleaner for this no? 
In the end, I think we should just not handle `else{...}` case. 

If we do not find it we do nothing and after the corresponding threshold, we clean the ongoing transfer. 

Wdyt ?

Victor-Salomon (commented):
Can you check my comment please about marking the transfer as Undefined ? 
Everything looks good otherwise 💸💸

PR #326, opened 2025-04-07
impr: add a transfer status retry to the tracking cleaner

Victor-Salomon:
This PR implements a retry mechanism before an outdated transfer is marked as undefined. The retryStatusVerification function is triggered based on the transfer direction. 

=> [Linear issue related](https://linear.app/velocitylabs/issue/VEL-346/fix-tx-tracking-unkown-setup)

Current status: On hold - Waiting for Ocelloids' historical aggregated data feature (should be released soon)


app/src/hooks/useOngoingTransferCleaner.ts
```diff
@@ -1,47 +1,95 @@
...
-          ...(explorerLink && { explorerLink }),
-        } satisfies CompletedTransfer)
-
-        addNotification({
-          message: 'Transfer tracking failed',
-          severity: NotificationSeverity.Warning,
-          dismissible: true,
-        })
-        captureException(new Error('Transfer tracking failed'), { extra: { ongoing } })
+    let cancelCleaning = false
```
NunoAlexandre, line 44: before reading the rest, no idea what this naming is for

app/src/hooks/useOngoingTransferCleaner.ts
```diff
@@ -1,47 +1,95 @@
...
+          if (retriedStatus === TxStatus.Undefined) {
+            addNotification({
+              message: 'Transfer tracking failed',
+              severity: NotificationSeverity.Warning,
+              dismissible: true,
+            })
+            captureException(new Error('Transfer tracking failed'), { extra: { ongoing } })
+          } else {
+            addNotification({
+              message: 'Transfer completed',
```
NunoAlexandre, line 80: What if `retriedStatus === TxStatus.Failed`?

app/src/hooks/useOngoingTransferCleaner.ts
Victor-Salomon: It's a [common React pattern](https://react.dev/reference/react/useEffect#fetching-data-with-effects) to prevent race conditions. Usually, the name is not very meaningful like the `ignore` or `cancel` as in the React Documentation.

But I'm all good to improve it. 
What do you think about `stopOngoingTransferCleaning`?

app/src/hooks/useOngoingTransferCleaner.ts
Victor-Salomon: I've not changed the core logic behind our terms. Completed transfers = Failed or Successful.
No matter if `retriedStatus === TxStatus.Failed` or `retriedStatus === TxStatus.Succeeded`, we register a completed transfer when we track a Bridge transfer for now. 

I agree we could adjust this logic in both places as this is not very accurate. I've already implemented this in getNotification() for Ocelloids XCM tracking. We could probably have a common helper or reuse it.

PR #330, opened 2025-04-09
chore: make sure that imports are in the right order

steoo:
Proposal and draft PR. 

At the moment we're not sorting the imports and it can get a bit messy – installed https://github.com/import-js/eslint-plugin-import/ as a dependency and added some rules. It's still in thorough testing because I want to make sure it's not messing with the named imports of specific modules. 

Also noticed this which actually is quite important:
Screenshot 2025-04-09 at 16 22 10

running `pnpm run lint` is catching some warnings, whilst the pre-commit `next lint` isn't. 

They're quite important warnings as they refer to missing dependencies which, in my experience, will always create problems of rendering consistency – if they're there on purpose, we should think of a fix.


jonidelv (approved):
lgtm 👍

NunoAlexandre (approved):

NunoAlexandre:
@steoo let's reload this and merge it once updated 👌

PR #335, opened 2025-04-13
Refactor useTransferForm dependencies for optimized reactivity

jonidelv:
While working on [this PR](https://github.com/velocitylabs-org/turtle/pull/332), I noticed a couple of improvements in the useTransferForm.ts file:
	•	Updated the dependency arrays in useCallback hooks to ensure consistent reactivity and avoid unnecessary re-renders.
	•	Adjusted the import path for formatAmount to use an absolute path, improving consistency across the codebase.

### DIFF
--- a/app/src/hooks/useTransferForm.ts
+++ b/app/src/hooks/useTransferForm.ts
@@ -17,7 +17,7 @@ import { switchChain } from '@wagmi/core'
 import { useCallback, useEffect, useState } from 'react'
 import { SubmitHandler, useForm, useWatch } from 'react-hook-form'
 import { mainnet } from 'viem/chains'
-import { formatAmount } from '../utils/transfer'
+import { formatAmount } from '@/utils/transfer'
 import useFees from './useFees'
 import useNotification from './useNotification'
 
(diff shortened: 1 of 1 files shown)


app/src/hooks/useTransferForm.ts
```diff
@@ -111,15 +111,7 @@ const useTransferForm = () => {
...
       isRouteAllowed(environment, destinationChain, sourceChain, tokenAmount)
     )
-  }, [
-    environment,
-    destinationChain,
-    sourceChain,
-    tokenAmount,
-    isValidating,
-    transferStatus,
-    tokenAmountError,
```
jonidelv, line 121: We were not using `tokenAmountError` within the allowFromToSwap callback function, so it’s better to remove it from the array.

app/src/hooks/useTransferForm.ts
```diff
@@ -162,7 +154,7 @@ const useTransferForm = () => {
       setValue('destinationChain', newValue)
       trigger()
     },
-    [setValue],
+    [setValue, trigger],
```
jonidelv, line 157: Trigger was missing in the dependency array even though we were using it inside the callback. This could be causing a bug.

steoo:
good catch! lgtm 👍🏼

steoo (approved):

PR #336, opened 2025-04-13
Fix animation and transition effects in ProgressBar

jonidelv:
While working on https://github.com/velocitylabs-org/turtle/pull/332, I noticed an improvement in the `ProgressBar` component. When the `outlinedProgressBar` prop was set to true, the animation (the width transition effect) was not functioning. This fix resolves that issue.

### DIFF
--- a/app/src/components/ProgressBar.tsx
+++ b/app/src/components/ProgressBar.tsx
@@ -21,10 +21,15 @@ export default function ProgressBar({ progress, outlinedProgressBar }: ProgressB
           />
         </div>
       ) : (
-        <div className="mb-4 h-2 rounded-full border border-turtle-secondary bg-white">
+        <div
+          className={cn(
+            'mb-4 h-2 rounded-full border border-turtle-secondary bg-white',
+            progress <= 0 && 'animate-pulse',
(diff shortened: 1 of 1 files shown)


app/src/components/ProgressBar.tsx
```diff
@@ -21,10 +21,15 @@ export default function ProgressBar({ progress, outlinedProgressBar }: ProgressB
           />
         </div>
       ) : (
-        <div className="mb-4 h-2 rounded-full border border-turtle-secondary bg-white">
+        <div
+          className={cn(
+            'mb-4 h-2 rounded-full border border-turtle-secondary bg-white',
+            progress <= 0 && 'animate-pulse',
```
jonidelv, line 27: using the same as here https://github.com/velocitylabs-org/turtle/blob/04ba41548bc85733e032097ed19c8c1f8416d4b4/app/src/components/ProgressBar.tsx#L15

NunoAlexandre:
Nice! I will let @Victor-Salomon give a final review 👌

Victor-Salomon (approved):
LGTM 🦾

PR #341, opened 2025-04-14
Conditional dividers in UI and wait for destination token

jonidelv:
I was going through the new swap UI and noticed a couple of things:

1. We’re currently showing the divider for the destination token, even though at that point we’re just displaying the token itself and not any amount. This initially made me think that a value (like the amount I’d receive) was going to appear there. Since nothing actually shows up, I think it might be better to remove the divider for the destination token to avoid any confusion.
2. I also noticed that the fee UI appears even before a destination token is selected. This can make the UI feel a bit off and harder to understand, especially when the token selection menu is still open. I think it would be a smoother experience if we waited for the destination token to be selected before showing the fee-related info.

Let me know what you think.

<p align="center">
  
  
</p>

**Note:**
It should be safe to merge into `feat/swaps-mvp`, even with the failing unit test. We’re already aware of the issue, and it’s also happening on that branch. If everything else looks good, we could consider force-merging it.

### DIFF
--- a/app/src/components/ChainTokenSelect.tsx
+++ b/app/src/components/ChainTokenSelect.tsx
@@ -49,6 +49,7 @@ interface ChainTokenSelectProps {
     sourceChainToDetermineOriginBanner: Chain | null
     error?: string
     clearable?: boolean
+    showVerticalDivider?: boolean
     orderBySelected?: boolean
     /** The token displayed at the top or below the selected token in the dropdown. Can be used to make it easier to select the source token again. */
     priorityToken?: Token | null
@@ -230,6 +231,7 @@ interface TokenAmountInputProps {
   token: {
     value: Token | null
     sourceChainToDetermineOriginBanner: Chain | null
+    showVerticalDivider?: boolean
   }
   amount?: {
     value: number | null
@@ -285,7 +287,7 @@ const TokenAmountInput = ({
             )}
           </div>
           <ChevronDown strokeWidth={0.2} className="ml-1" />
-          <VerticalDivider />
+          {token.showVerticalDivider && <VerticalDivider />}
           <div className="align-center ml-1 flex flex-col">
             <input
               data-cy="amount-input"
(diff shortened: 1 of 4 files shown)


app/src/components/ChainTokenSelect.tsx
```diff
@@ -49,6 +49,7 @@ interface ChainTokenSelectProps {
     sourceChainToDetermineOriginBanner: Chain | null
     error?: string
     clearable?: boolean
+    showVerticalDivider?: boolean
```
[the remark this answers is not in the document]
jonidelv: Sounds good. I changed it to check if we are disabling the amount, meaning that if the amount is not enabled, the input does not accept a value, and thus, the separator shouldn't show.

app/src/components/Transfer.tsx
```diff
@@ -146,7 +146,7 @@ export default function Transfer() {
   else if (balanceData?.value === 0n) amountPlaceholder = 'No balance'
   else amountPlaceholder = formatAmount(Number(balanceData?.formatted), 'Longer')
 
-  let receiveAmountPlaceholder = 'Receive Amount'
+  let receiveAmountPlaceholder = 'Receive Token'
```
jonidelv, line 149: We are not showing the amount in the receiving token input, only the token itself.

NunoAlexandre: This should be 'amount', not token  👀 

```suggestion
  let receiveAmountPlaceholder = 'Receive amount'
```

noahjoeris:
Is it compliant with #346 ?
We now always show a receive amount

jonidelv:
It should be, since I'm using this: `!!amount?.value || !!amount?.placeholder`, meaning that if we don't have a value or a placeholder, that space is empty (blank), so there's no point in showing the divider.

NunoAlexandre (approved):
LGTM :)

PR #346, opened 2025-04-15
impr: show receive amount for transfers

noahjoeris:
This PR adds the receive amount also for transfers. It deducts the fees if there are some.


Victor-Salomon (approved):
Nice 🦾

it would have been nice to create some types for the isSwap, isSameChainSwap and isSwapWithTransfer instead of repeating the extended object, but we never use those anywhere else

PR #352, opened 2025-04-16
impr: make address copyable and adjust dropdown trigger area

noahjoeris:
This PR adds copying the wallet address. And adjusts the area to click to open the dropdown. As discovered in our feature testing.


NunoAlexandre:
@noahjoeris the actually clickable area feels right now but the mouse cursor still shows as if it's clickable everywhere. Try to select a chain & token and then place the mouse next to the "Connect" button to reproduce what I mean.

noahjoeris:
@NunoAlexandre That's the same behavior as on main. I replicated it as you said.

noahjoeris:
I can fix that in another PR

NunoAlexandre:
> @NunoAlexandre That's the same behavior as on main. I replicated it as you said.

You are right apparently, but clearly it's wrong on main from a UX perspective. And what I meant to ask to replicate is the fix of the clickable area 🙏

NunoAlexandre:
> I can fix that in another PR

Can we keep some common sense on opening new PRs? There is no need to open a new PR for every review comment, specially when it's so interconnected. Thanks

noahjoeris:
Sure. But your comment was too late, I had already merged.

PR #353, opened 2025-04-17
fix: remove sentry capture

noahjoeris:
Removes sentry capture if a token is not found via multilocation. Which means we do not support it via our registry.


NunoAlexandre (approved):
Fair 👍

Victor-Salomon (approved):
Good to me ;)

PR #354, opened 2025-04-17
fix: wrong origin chain banner

noahjoeris:
This small PR fixes the issue of showing a wrong origin banner on a swap

image


NunoAlexandre (approved):
Voila ✨

PR #360, opened 2025-04-22
Avoid stylin leaking in the Widget

steoo:
[Linear Issue](https://linear.app/velocitylabs/issue/VEL-382/widgetfix-css-variables-over-ride-some-of-the-host-app-css-config)

The issue we currently have is that the widget, being it a Vite app, is loading and injecting styles. This means that it will: 

- Leak styles in the consumer's app
- Suffer from styles that are in the consumers app

What has been done: 
- Made the package names consistent `@velocitylabs-org/turtle-<package-name>`
- Generalised eslint config for ui and widget. The idea is to have one config for all, but there are some inconsistencies between vite, next and non-framework packages.
- Removed HeroUI from the Widget
- Removed Tooltip and Button implementation from Widget
- Removed the Dazzed font
- Created the UI Package that includes Button and Tooltip 
- Implemented scripts from the root package.json to run on all the packages that need to run that specific script
- Small lint issues

Some notes and next steps:
- One curious thing is that some lines of code from app/widget/site are failing the lint check, yet they were already there and never failed. Gotta look into it. 
- Replace Button and Tooltip from App as well, but I thought Widget was a good place where we could test this.
- Publish the Widget with the updated styles / packages
- Share as much config (eslint, prettier, typescript) between as many packages as possible.


Victor-Salomon:
@steoo, do we still want to implement this shadow DOM if it's not solving the issue?

steoo:
@Victor-Salomon, no, I'm pushing a different solution 🫠

NunoAlexandre (approved):
I don't find anything that seems off 👍

app/src/hooks/useTransferForm.ts
```diff
@@ -51,7 +51,7 @@ const useTransferForm = () => {
     trigger,
     formState: { errors, isValid: isValidZodSchema, isValidating },
   } = useForm<FormInputs>({
-    resolver: zodResolver(schema),
+    resolver: zodResolver(schema as any),
```
Victor-Salomon, line 54: I had to do it on the widget as well. 🥴

steoo: I wonder how did it pass in the first place 🤔 it's not the first one I find

packages/ui/index.tsx
```diff
@@ -0,0 +1,4 @@
+export * from './src/components/LoadingIcon'
+export * from './src/components/Button'
+export * from './src/components/Tooltip'
+export * from './src/types/global'
```
Victor-Salomon, line 4: I'm just highlighting the warning provided by Github Actions.

packages/ui/src/components/Button.tsx
```diff
@@ -0,0 +1,99 @@
...
+        <div className="flex items-center">
+          {icon && <span className="mr-3">{icon}</span>}
+          <span className="mr-3">{label}</span>
+        </div>
+      )}
+
+      {!loading && (children || label)}
+    </button>
+  )
+}
```
Victor-Salomon, line 99: 🦾

widget/src/components/ChainSelect.tsx
```diff
@@ -8,11 +8,10 @@ import { useEnsAvatar } from 'wagmi'
 import { normalize } from 'viem/ens'
 import { cn } from '@/utils/helper'
 import ChainIcon from '@/assets/svg/ChainIcon'
-import { Tooltip } from './Tooltip'
+import { Button, Tooltip } from '@velocitylabs-org/turtle-ui'
```
Victor-Salomon, line 11: 🤩

widget/src/App.tsx
```diff
@@ -0,0 +1,14 @@
...
+function App() {
+  return (
+    <Providers>
+      <Widget />
+      <SubstrateWalletModal />
+    </Providers>
+  )
+}
+
+export default App
```
Victor-Salomon, line 14: I'm not sure to understand why we need this App file?

steoo: I think it was there but I will wipe it out of existence if not needed

Victor-Salomon (approved):
@steoo everything looks good🦾
Nice work mate. 
I've left a comment about the `widget/src/App.tsx` file.

app/src/hooks/useTransferForm.ts
Victor-Salomon: 🤷‍♂️

widget/src/App.tsx
Victor-Salomon: I would say, the `main.tsx` file outside of the `src` folder should do the same job. Having it out avoids exporting in the Widget package. 

Could you confirm you can run the project locally without this App file?

widget/tailwind.config.ts
```diff
@@ -99,7 +98,7 @@ const config: import('tailwindcss').Config = {
     },
   },
   darkMode: ['class'],
-  plugins: [heroui(), tailwindcssAnimate, tailwindcssMotion],
```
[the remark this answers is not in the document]
steoo: lol

widget/src/App.tsx
steoo: checked and apparently useless – deleting

PR #361, opened 2025-04-23
Refine exceedsTransferableBalance to account bridging 

jonidelv:

PR #364, opened 2025-04-24
Fix: Refactor fee calculation logic in setTransferableBalance

jonidelv:
In [yesterday's PR ](https://github.com/velocitylabs-org/turtle/pull/361), this logic needed to be refactored as well. Doing it now.


noahjoeris (approved):
looks good 👍

PR #367, opened 2025-05-01
Fix UX when changing source chain/token with a manual recipient set

NunoAlexandre:
Fixes https://linear.app/velocitylabs/issue/VEL-393/bug-manual-recipient-issue-when-updating-a-source-token

## Changes

- [x] Fix & improve styling
- [x] Wipe manual recipient when changing source chain/token
   We can decide to keep it if the new source+dest chains are compatible but let's keep it simple for now

### Before

https://github.com/user-attachments/assets/276f0d02-46cb-46d9-a6a5-e6350434f4da

### After

https://github.com/user-attachments/assets/89f94fef-4f60-41ec-9ad4-49c96d029e3a

### DIFF
--- a/app/src/components/ChainTrigger.tsx
+++ b/app/src/components/ChainTrigger.tsx
@@ -78,7 +78,7 @@ export default function ChainTrigger({
       >
         <div className="flex h-[3.5rem] flex-grow items-center gap-1">
           <div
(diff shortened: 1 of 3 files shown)


jonidelv (approved):
lgtm

steoo (approved):

PR #369, opened 2025-05-06
fix: stable coin balance fix + paraspell SDK upgrade

Victor-Salomon:
This PR upgrade the PS SDK and fix the USDC/T balances issues. 
It also update the Dryrun with the origin verification as requested by Dudo. 

image


app/src/utils/paraspellTransfer.ts
```diff
@@ -47,7 +46,7 @@ export const createTransferTx = async (
     .from(sourceChainFromId as TNodeDotKsmWithRelayChains)
     .to(destinationChainFromId)
     .currency({ ...currencyId, amount: sourceAmount })
-    .address(recipient, destinationChainFromId === 'Ethereum' ? getAccountId32(sender.address) : '')
```
Victor-Salomon, line 50: @NunoAlexandre I want to confirm with Dudo if I need to pass _`.senderAddress(sender.address)`_ to the Builder. 
The documentation is unclear.

app/src/utils/paraspellTransfer.ts
```diff
@@ -47,6 +48,9 @@ export const createTransferTx = async (
     .to(destinationChainFromId)
     .currency({ ...currencyId, amount: sourceAmount })
     .address(recipient)
+    .senderAddress(
+      destinationChainFromId === 'Ethereum' ? getAccountId32(sender.address) : sender.address,
```
NunoAlexandre, line 52: I don't follow this 👀 

1. Having "Destination" and "From" in the same name is confusing. I know it was there before but we should improve that
2. If the chain is Ethereum, why would we get the AccountId32? Shouldn't it be the way around, given that in Ethereum addresses have 20 bytes and not 32? In either case, this is probably not a good idea any ways because for chains like Moonbeam we need to handle EVM-like addresses. So we shouldn't have a rule of thumb like this

PR #370, opened 2025-05-06
Scope CSS for Widget

steoo:
[Linear Issue](https://linear.app/velocitylabs/issue/VEL-397/scope-widget-css)

The widget CSS has been scoped and a reset was applied to the best we can do right now. 
From the perspective of a Consumer, the widget will have to be preferably placed in a container that doesn't pollute its styling, eg: 
- as a direct child of `<body>`
- in a portal

The reset makes sure that we have the font to be the system one and keeps as much consistency as possible. 

This solution will still inject tailwind into the consumer app – one solution would be to either obfuscate or prefix our classes, but it's not a quick fix, or at least not one I found. I'd rather let people test this thing first and then we can tackle it, if it's an issue.


Victor-Salomon (approved):
Code looks good to me 🦾

PR #372, opened 2025-05-07
feat: Add analytics dashboard

jonidelv:
This PR introduces a new analytics project that provides enhanced visualization and control over our analytics data.

### Key Features
Complete control over analytics data display
Isolated implementation that doesn't affect other projects
Next.js frontend with dedicated API routes
MongoDB integration for advanced querying and aggregation capabilities

### Technical Details
Frontend: Next.js with dedicated routes
Backend: API routes within Next.js that communicate with MongoDB
Database: MongoDB with Mongoose wrapper for improved structure
Developer Experience: Maintains consistency with the main app, using shared packages like registry to avoid code duplication
Performance: Leverages React Compiler, which automatically handles optimization primitives (useMemo, useCallback, memo, etc.)

### Dependencies
Utilizes shared packages for code reuse
Will integrate the shared assets module (icons and chain images) once available

### Current Status
⚠️ The implementation is ready for review but once the shared assets module is available, I'll add it in a subsequent update.


analytics/tsconfig.json
steoo, line 1: this can already leverage the shared tsconfig tho

steoo (approved):

jonidelv:
Yes, let me share the `env` file I'm using so you guys can check it out in development as well.

jonidelv:
This is the PR live preview [turtle-analytics-git-feat-analytics-dashboard-velocity-labs.vercel.app](https://turtle-analytics-git-feat-analytics-dashboard-velocity-labs.vercel.app/)

noahjoeris:
Oh wow this looks stunning!

analytics/src/constants/index.tsx
```diff
@@ -0,0 +1,25 @@
+import { Chain, Token, REGISTRY, Ethereum, RelayChain } from '@velocitylabs-org/turtle-registry'
```
Victor-Salomon, line 1: 😍

Victor-Salomon (approved):
Everything looks solid! There's a lot of code to review, but overall it seems fine, and the design is stunning 🔥🤩.

I left a comment about the manual props validation in the API route. Using Zod could be a nice alternative there.

General comments: 
I noticed the pattern of using API routes with React Query. While that works well, another approach could be to use Server Components and perform DB calls directly in the pages. No need to refactor now, though as it would be quite heavy.

Lastly, it might be worth adding support for dynamic params (with pagination) in the API requests in the future. (ex: if we plan to query beyond a top 5)

Impressive job overall @jonidelv 👏 💥🦾

analytics/src/app/api/create-transaction/route.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (!validateRequest(request)) {
+      return new Response(JSON.stringify({ message: 'Forbidden' }), {
+        status: 403,
+      })
+    }
+
+    await dbConnect()
+    const data = await request.json()
+
+    const requiredFields = [
```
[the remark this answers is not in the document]
NunoAlexandre: @jonidelv what does your reaction to Victor here means? Will you address it before merging or afterwards?

analytics/src/app/api/create-transaction/route.ts
```diff
@@ -0,0 +1,77 @@
...
+        { status: 409 }, // 409 Conflict status code
+      )
+    }
+
+    const transaction = new Transaction(data)
+    await transaction.save()
+
+    return NextResponse.json(
+      {
+        message: 'Transaction created successfully',
```
NunoAlexandre, line 68:
```suggestion
        message: 'Transaction stored successfully',
```

NunoAlexandre: We should change the "create" terminology to "track" or "store".

analytics/src/app/api/simple-summary/route.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    const totalVolumeUsd = volumeResult[0]?.total || 0
+
+    return NextResponse.json({
+      totalVolumeUsd,
+      totalTransactions,
+    })
+  } catch (error) {
+    console.error('Database error:', error)
+    return NextResponse.json({ error: 'Internal Server Error' }, { status: 500 })
```
NunoAlexandre, line 41: We probably should track this with Sentry

analytics/src/app/api/summary/route.ts
```diff
@@ -0,0 +1,134 @@
...
+
+      // Successful transactions count
+      Transaction.countDocuments({ status: 'succeeded' }),
+
+      // Last 5 transactions
+      Transaction.find()
+        .sort({ txDate: -1 })
+        .limit(5)
+        .select(
+          'txDate sourceTokenId sourceTokenSymbol sourceTokenAmount sourceTokenAmountUsd sourceChainUid sourceChainName destinationTokenId destinationTokenSymbol destinationChainUid destinationChainName status',
```
NunoAlexandre, line 46: What happens if one of these fields' names change or is deleted? do we get an error somewhere or will it still go through silently?

analytics/tsconfig.json
jonidelv: Yes, I have implemented that.

analytics/src/app/api/create-transaction/route.ts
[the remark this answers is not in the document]
jonidelv: Sorry, I agree with that. I pushed an implementation of Zod for this

analytics/src/app/api/simple-summary/route.ts
jonidelv: Done, created a project in Sentry.

analytics/src/app/api/summary/route.ts
jonidelv: Good point, just updated this to be more robust, instead of simple string importing the namespaces from the model

jonidelv:
@Victor-Salomon Thanks for the feedback! I've implemented Zod for validation as suggested. There's also an option to leverage Zod within mongoose models that I'm planning to explore further after deployment.

Regarding Server Components vs. API routes: you're right that we could fetch data directly on the backend. I opted for frontend fetching because it allows us to render the page immediately while showing loading indicators for data. Since we don't need SEO indexing for this content, this approach works well for user experience, but I'm open to revisiting this in the future.

For now, we're only displaying the latest 5 items, but I have a call scheduled with Nico to discuss what data he'd like to see, which might necessitate adding pagination support.

Victor-Salomon:
@jonidelv amazing! 
Thanks a lot for the extra effort to implement ZOD 🦾

PR #373, opened 2025-05-07
started migrating registry, icons, models, and utils to monorepo packages

steoo:


jonidelv (approved):
lgtm 👏

packages/registry/src/types/select.ts
```diff
@@ -0,0 +1,32 @@
+import { ReactNode } from 'react'
```
Victor-Salomon, line 1: Are we sure about moving this `types/select` file inside the registry package?

packages/registry/tsconfig.tsbuildinfo
```diff
@@ -0,0 +1 @@
...,450],[672,450],[427,450],[429,451],[431,452],[488,453],[432,450],[495,454],[433,455],[702,456],[435,457],[434,458],[438,439],[523,459],[522,460],[512,452],[514,450],[515,461],[513,462],[516,463],[506,464],[517,465],[474,466],[473,467],[466,468],[478,469],[550,470],[467,471],[545,472],[554,473],[477,474],[558,475],[468,475],[489,476],[708,477],[439,478],[490,479],[709,478],[448,478],[705,480],[532,481],[703,482],[706,480],[493,483],[704,482],[491,484],[492,485],[694,486],[695,487],[457,4],[481,488],[576,471],[579,479],[413,479],[540,479],[414,479],[650,479],[651,489],[415,490],[655,491],[675,492],[569,493],[574,494],[689,495],[676,289],[690,496],[677,289],[425,497],[678,498],[679,499],[681,500],[682,501],[684,502],[691,497],[680,503],[683,492],[707,504],[686,505],[687,506],[688,507],[418,389],[587,508],[508,509],[582,479],[469,510],[459,511],[674,512],[444,513],[713,514],[460,515],[461,516],[573,517],[700,518],[701,518],[699,479],[711,519],[712,519],[710,479],[56,520],[937,521],[945,522],[938,523],[942,524],[941,525],[940,526],[939,527],[934,520],[933,528],[936,529],[946,530],[947,4],[935,531],[944,532],[943,533]],"latestChangedDtsFile":"./dist/types/select.d.ts","version":"5.8.3"}
```
Victor-Salomon, line 1: Just to confirm (my own understanding), what's the purpose of this file?

Victor-Salomon (approved):
It looks good but I've not tested it. 
I left a couple of comments.

packages/registry/tsconfig.tsbuildinfo
steoo: it's trash that needs to be ignored and it keeps on appearing 😠

packages/registry/src/types/select.ts
steoo: uhm good point, yet I would argue they shouldn't be in `models` in the first place, as they're shaping props of Components rather than models. I see that `SelectProps` is not even used :/

PR #375, opened 2025-05-09
Support MYTH transfers between Mythos and AssetHub

NunoAlexandre:
* Mythos-native MYTH


Victor-Salomon (approved):
Looks all good here 🦾

PR #376, opened 2025-05-09
Fix: Extend Metrics Support to Swaps and Polkadot Network

jonidelv:
Prior to this change, we only tracked Moonbeam token transfers on the Polkadot network. This PR:
	•	Extends transfer tracking to all asset transfers across the Polkadot ecosystem (including swaps).
	•	Adds the initial plumbing required for our upcoming analytics deployment.


app/src/hooks/useParaspellApi.ts
```diff
@@ -100,24 +99,8 @@ const useParaspellApi = () => {
       params.onComplete?.()
     })
 
-    // We intentionally track the transfer on submit. The intention was clear, and if it fails somehow we see it in sentry and fix it.
-    if (params.environment === Environment.Mainnet && isProduction) {
```
jonidelv, line 104: the `trackTransferMetrics` already checks for `params.environment === Environment.Mainnet && isProduction`

app/src/hooks/useParaspellApi.ts
```diff
@@ -187,6 +171,8 @@ const useParaspellApi = () => {
         } catch (error) {
           handleSendError(params.sender, error, setStatus, event.txHash.toString())
         }
+
+        sendMetrics({ params, txId: event.txHash?.toString() ?? '', senderAddress, sourceTokenUSDValue, destinationTokenUSDValue, date })
```
NunoAlexandre, line 175: This doesn't seem like a good idea. Can we generate a unique string in case txHash is null/undefined?
Referring to `txId: event.txHash?.toString() ?? ''`

noahjoeris: Hmm if the txHash is not defined we shouldn't track it at all no? It's not a valid transaction! It was not successful.

noahjoeris (commented):
Looks good! Left one comment to your conversation

app/src/hooks/useParaspellApi.ts
jonidelv: The previous implementation allowed saving a transfer to analytics data even without a txHash. I believe it’s better to discard the transfer if it doesn’t include a txHash, but I aimed to maintain compatibility with the existing behavior.
Let me know how you’d like to proceed, I can discard transfers entirely if the txHash is missing.

noahjoeris: For me it makes sense to discard them

jonidelv: ok, just pushed the fix 👌

app/src/hooks/useParaspellApi.ts
```diff
@@ -400,3 +408,41 @@ const useParaspellApi = () => {
 }
 
 export default useParaspellApi
+
+interface SendMetricsParams {
+  params: TransferParams
+  txId: string
```
noahjoeris, line 414: Let's better do this
```suggestion
  txId?: string
```

And then you don't need to do `... ?? ''`

noahjoeris (approved):
Left one small comment about the design. But besides that it looks good to merge.

app/src/hooks/useParaspellApi.ts
```diff
@@ -400,3 +408,41 @@ const useParaspellApi = () => {
...
+
+function sendMetrics({
+  params,
+  txId,
+  senderAddress,
+  sourceTokenUSDValue,
+  destinationTokenUSDValue,
+  date,
+}: SendMetricsParams) {
+  if (!txId) return
```
Victor-Salomon, line 429: Why not verify the txId at a lower level, in `trackTransferMetrics`, to avoid creating this sendMetrics wrapper, and SendMetricsParams type?

```
export async function trackTransferMetrics(data: TransferMetric) {
  if (
    !process.env.NEXT_PUBLIC_FIREBASE_PROJECT_ID ||
    data.environment !== Environment.Mainnet ||
    !isProduction ||
    !id
  ) {
    return
  }
```

Victor-Salomon (approved):
It looks good 🦾
I've left one comment

PR #379, opened 2025-05-12
Use package/registry in all the apps

steoo:


NunoAlexandre:
@steoo it's weird that the diff of this PR includes the latest PR I merged adding support to transfer MYTH from Mythos <> AH 👀

widget/src/components/ChainTokenSelect.tsx
```diff
@@ -366,7 +362,7 @@ const ChainList = ({
             >
               <div className="flex items-center gap-2">
                 <img
-                  src={option.logoURI}
+                  src={option.logoURI as string}
```
NunoAlexandre, line 365: do you remember why this was needed?

packages/registry/src/mainnet/tokens.ts
```diff
@@ -1,13 +1,39 @@
...
+import shibLogo from "@velocitylabs-org/turtle-assets/logos/shib.svg";
+import tbtcLogo from "@velocitylabs-org/turtle-assets/logos/tbtc.svg";
+import tonLogo from "@velocitylabs-org/turtle-assets/logos/ton.svg";
+import usdcLogo from "@velocitylabs-org/turtle-assets/logos/usdc.svg";
+import usdtLogo from "@velocitylabs-org/turtle-assets/logos/usdt.svg";
+import vdotLogo from "@velocitylabs-org/turtle-assets/logos/vdot.svg";
+import vethLogo from "@velocitylabs-org/turtle-assets/logos/veth.svg";
+import wbtcLogo from "@velocitylabs-org/turtle-assets/logos/wbtc.svg";
+import wethLogo from "@velocitylabs-org/turtle-assets/logos/weth.svg";
+import wstethLogo from "@velocitylabs-org/turtle-assets/logos/wsteth.svg";
```
NunoAlexandre, line 25: imo this change was unnecessary and just adds extra noise to review this pr

NunoAlexandre (approved):
An important step forward. Thanks for moving this big needle, @steoo ✌️

widget/src/components/ChainTokenSelect.tsx
steoo: nextjs and vite handle imports differently – therefore for next.js `option.logoUri` will be an object (with an `src`) whilst for vite (widget) it will be a string – so atm te logoURI is defined as `string | Record<string, any>` and needs to be casted atm

packages/registry/src/mainnet/tokens.ts
steoo: this is unfortunately inevitable because the `Chain` and `Token`, which are supposed to be models (and therefore data only) are also mapping the logos. I don't love the idea that they're there but that's how it was made – there are other ways of doing this but I think it was out of scope to solve it in this PR, but if I didn't apply this middle change, the logos wouldn't have worked in either the App or Widget

jonidelv (approved):
Good job, gonna use this in Analytics 👏

app/src/components/Transfer.tsx
```diff
@@ -9,12 +11,8 @@ import useEthForWEthSwap from '@/hooks/useEthForWEthSwap'
 import useSnowbridgeContext from '@/hooks/useSnowbridgeContext'
 import useTransferForm from '@/hooks/useTransferForm'
 import { WalletInfo } from '@/hooks/useWallet'
-import { TokenAmount } from '@/models/select'
-import { Hydration } from '@/registry/mainnet/chains'
-import { EthereumTokens } from '@/registry/mainnet/tokens'
-import { Balance } from '@/services/balance'
+
```
noahjoeris, line 14: Is there a reason for this empy line? I usually remove it again because it annoys me 😆

steoo:
```suggestion
```

package.json
```diff
@@ -10,9 +10,13 @@
     "start": "turbo run start",
     "test": "turbo run test",
     "watch": "turbo run watch",
-    "changeset": "changeset"
+    "changeset": "changeset",
+    "prebuild": "turbo run prebuild",
+    "knip": "knip"
```
Victor-Salomon, line 15: @steoo what is `knip` script used for? 
And when should we use `prebuild` script? 
Thanks :)

packages/registry/package.json
```diff
@@ -7,8 +7,9 @@
   "type": "module",
   "author": "Stefano Imparato <stefano@velocitylabs.org>",
   "scripts": {
-    "build": "tsc -b",
-    "watch": "tsc --watch",
+    "prebuild": "node script/generateEnv.cjs",
```
Victor-Salomon, line 10: @steoo I see it now

packages/registry/src/generated-env.ts
```diff
@@ -0,0 +1 @@
+export const DWELLIR_KEY = "4face3c5-860f-474e-84f1-8b0fbd203803";
```
Victor-Salomon, line 1: Is it normal the key is hard coded? 
I understand it was generated with prebuild.

packages/registry/src/mainnet/tokens.ts
Victor-Salomon: I had to do the same with the Widget... 🫠

Victor-Salomon (approved):
I've left a couple of comments for my understanding but I've not seen any wrong change! 

This one was SAVAAAGE 💥
Nice job @steoo 🦾

packages/registry/src/types/chain.ts
```diff
@@ -4,7 +4,7 @@ export interface Chain {
   /** Full name of the chain. */
   name: string;
   /** image URL of the chain logo. */
-  logoURI: string;
+  logoURI: string | { [key: string]: string };
```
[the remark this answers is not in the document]
steoo: we do indeed

packages/registry/src/generated-env.ts
steoo: it should indeed being git ignored and generated on builds only, good point

package.json
steoo: knip is a utility for checking unused and mess in your code, quite nice!
prebuild yes, we keep!

PR #380, opened 2025-05-16
chore: update paraspell to v9.2.2

noahjoeris:


app/src/hooks/useParaspellApi.ts
```diff
@@ -213,85 +202,90 @@ const useParaspellApi = () => {
 
     const routerPlan = await createRouterPlan(params)
 
-    const step1 = routerPlan.at(0)
-    if (!step1) throw new Error('No steps in router plan')
+    const firstTransaction = routerPlan.at(0)
```
Victor-Salomon, line 205: Definitely better than `step1`, but in the end, it will be the last transaction, no? send or bridge + swap + send

Victor-Salomon (approved):
Bingo 🤠
All good to me. 
I made a notification on a variable naming. But for now, I guess it's good since we don't support yet more than swap + transfer.

app/src/hooks/useParaspellApi.ts
noahjoeris: There could be a 2nd transaction. So we access the first one via `routerPlan.at(0)`. But yeah due to our filtering of routes there should only be 1 tx so far in the array. But not anymore with the upcoming swap changes.

PR #383, opened 2025-05-21
Fix logos not showing in Ongoing and Completed transfers

steoo:
The transfers are being saved with the entirety of the Chain and/or Token in the storage itself – this is an issue, especially for images that might change paths (like logoURI). Therefore now we map the Chain (or Token) to the registry where the most up-to-date information relies. 

It's not the ideal solution because I would like the mapping to happen on a higher level, but for the sake of being quick this is a good short term fix. I would need to investigare more on how the transfers are mapped in the first, but I feel it would be a potential breaking change rn which is out of scope.

### DIFF
--- a/app/next.config.mjs
+++ b/app/next.config.mjs
@@ -54,7 +54,6 @@ const nextConfig = {
       },
     ],
   },
-  transpilePackages: ['@velocitylabs-org/turtle-ui', '@velocitylabs-org/turtle-registry'],
 }
 
 // Sentry is enabled only in production builds to ensure error tracking in the live environment.
--- a/app/next.config.mjs
+++ b/app/next.config.mjs
@@ -1,6 +1,6 @@
 /** @type {import('next').NextConfig} */
-import { withSentryConfig } from '@sentry/nextjs'
 import withBundleAnalyzer from '@next/bundle-analyzer'
+import { withSentryConfig } from '@sentry/nextjs'
 
 const isDevelopment = process.env.NODE_ENV === 'development'
 const vercelDomain = process.env.NEXT_PUBLIC_VERCEL_URL
@@ -54,6 +54,7 @@ const nextConfig = {
       },
     ],
   },
+  transpilePackages: ['@velocitylabs-org/turtle-ui', '@velocitylabs-org/turtle-registry'],
 }
 
 // Sentry is enabled only in production builds to ensure error tracking in the live environment.
--- a/app/src/components/completed/TransactionHistory.tsx
+++ b/app/src/components/completed/TransactionHistory.tsx
@@ -1,3 +1,4 @@
+import { Chain, MainnetRegistry, Token } from '@velocitylabs-org/turtle-registry'
 import { cn } from '@velocitylabs-org/turtle-ui'
 import { CompletedTransfer } from '@/models/transfer'
 import { formatCompletedTransferDate } from '@/utils/datetime'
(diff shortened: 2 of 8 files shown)


NunoAlexandre:
> The transfers are being saved with the entirety of the Chain and/or Token in the storage itself – this is an issue, especially for images that might change paths (like logoURI). Therefore now we map the Chain (or Token) to the registry where the most up-to-date information relies.
> 
> It's not the ideal solution because I would like the mapping to happen on a higher level, but for the sake of being quick this is a good short term fix. I would need to investigare more on how the transfers are mapped in the first, but I feel it would be a potential breaking change rn which is out of scope.

Let me explain the context that makes your catch a good one :)

Initially, we pointed to logos hosted on external services. And we stored the whole chain/token data because both chain and token metadata might change. For example, the decimals of a token _can_ change, as well as their logo, symbol, etc. It's an unusual or rare occasion but it can happen. Most importantly, we didn't want to just store our own token `id` and use thats to index / look up the token data from our registry because we were well aware that the `id` we were using could very well change in the then near future and because we were planning on storing the txs in the local storage only for the time being.  And thus for us it made sense to just store the whole data set and basically freeze a snapshot of a transfer at the moment it was made.

In the meantime I designed and compiled a standard token & chain logo set for Turtle that lives within the project (ie., we broke the external link dependency) which indeed introduced the room for broken links as you just experienced and pointed out. I oversaw that.

I will review the PR tomorrow 🙏

jonidelv (approved):
Looks good just some thoughts regarding performance:
For every transfer/transaction, we're doing 4 separate .find() operations, this creates overhead, especially since these operations run on every render even when transfers or ongoingTransfers haven't changed since they are not memoized.
How about creating lookup maps outside the component to avoid repeated computations:

```javascript
// Outside component or in useMemo
const chainsByUid = MainnetRegistry.chains.reduce<Record<string, Chain>>(
  (acc, chain) => {
    acc[chain.uid] = chain;
    return acc;
  }, 
  {}
);

const tokensById = MainnetRegistry.tokens.reduce<Record<string, Token>>(
  (acc, token) => {
    acc[token.id] = token;
    return acc;
  }, 
  {}
);

// Then in the map:
{transfers.reverse().map((tx, idx) => {
  const enrichedTx = {
    ...tx,
    sourceToken: tokensById[tx.sourceToken.id],
    destinationToken: tokensById[tx.destinationToken?.id],
    sourceChain: chainsByUid[tx.sourceChain.uid],
    destChain: chainsByUid[tx.destChain.uid]
  };
  return <TransactionDialog key={idx + tx.id + tx.sender} tx={enrichedTx} />;
})}
```

steoo:
> Looks good just some thoughts regarding performance: For every transfer/transaction, we're doing 4 separate .find() operations, this creates overhead, especially since these operations run on every render even when transfers or ongoingTransfers haven't changed since they are not memoized. How about creating lookup maps outside the component to avoid repeated computations:
> 
> ```js
> // Outside component or in useMemo
> const chainsByUid = MainnetRegistry.chains.reduce<Record<string, Chain>>(
>   (acc, chain) => {
>     acc[chain.uid] = chain;
>     return acc;
>   }, 
>   {}
> );
> 
> const tokensById = MainnetRegistry.tokens.reduce<Record<string, Token>>(
>   (acc, token) => {
>     acc[token.id] = token;
>     return acc;
>   }, 
>   {}
> );
> 
> // Then in the map:
> {transfers.reverse().map((tx, idx) => {
>   const enrichedTx = {
>     ...tx,
>     sourceToken: tokensById[tx.sourceToken.id],
>     destinationToken: tokensById[tx.destinationToken?.id],
>     sourceChain: chainsByUid[tx.sourceChain.uid],
>     destChain: chainsByUid[tx.destChain.uid]
>   };
>   return <TransactionDialog key={idx + tx.id + tx.sender} tx={enrichedTx} />;
> })}
> ```

love this, thanks, will do

Victor-Salomon (approved):
Looks good, and even better with @jonidelv suggestion 🦾

PR #387, opened 2025-05-22
Fix: Simplify request validation logic.

jonidelv:
We don't need this logic because the browser is already validating CORS.


Victor-Salomon (approved):
🧹🧹

steoo (approved):

PR #388, opened 2025-05-23
Chore: Enhance analytics security using server actions

jonidelv:
This PR improves the overall security of the analytics feature by implementing the following changes:

- Migrated API routes to server actions: Server actions are not exposed as traditional API endpoints, which helps prevent unauthorized external access. Communication between the frontend and server actions is securely handled by Vercel.

- Token security: The analytics API is now called from the server using server actions, eliminating the need to expose the auth token in the frontend (avoiding the use of NEXT_PUBLIC_ variables, which would make the token visible in client side code).


analytics/README.md
```diff
@@ -0,0 +1,89 @@
...
+   ```
+5. Linter and format can be run in this directory `/analytics`
+
+## API Endpoints
+
+- **POST /api/create-transaction**: Submit new transaction data
+- **GET /api/simple-summary**: Get basic analytics data (total volume and transaction count) 
+  - This endpoint is used for displaying transaction data in the public landing page of Turtle
+
+### Security Measures for Transaction Submission
```
jonidelv, line 68: What do you think about including this here?

analytics/src/app/layout.tsx
```diff
@@ -10,7 +10,7 @@ export default function RootLayout({
 }>) {
   return (
     <html lang="en" className="bg-muted/60">
-      <body>
+      <body suppressHydrationWarning>
```
jonidelv, line 13: Some browser extensions add attributes to the body tag, which causes a mismatch between the client body and the server's response, resulting in warnings. This helps eliminate those warnings.

analytics/README.md
```diff
@@ -0,0 +1,89 @@
...
+2. Install dependencies in the root directory (see root directory for instructions since this is a monorepo handled with Turbo, where you can learn how to add/remove dependencies and run projects):
+   ```
+   pnpm install
+   ```
+3. Copy `.env.local.example` to `.env.local` and fill in the required environment variables
+4. Run the development server in root directory:
+   ```
+   pnpm run dev-analytics
+   ```
+5. Linter and format can be run in this directory `/analytics`
```
steoo, line 60: ?

analytics/src/app/api/simple-summary/route.ts
```diff
@@ -33,12 +59,18 @@ export async function GET(request: Request) {
...
+      NextResponse.json({
+        totalVolumeUsd,
+        totalTransactions,
+      }),
+      origin,
+    )
   } catch (error) {
-    captureServerError(error as Error)
-    return NextResponse.json({ error: 'Internal Server Error' }, { status: 500 })
+    await captureServerError(error as Error)
```
steoo, line 70: I've seen this repeated some times, would it be valuable to create a wrapper that handles the `try catch` always the same way and extend/use that?

analytics/src/models/Transaction.ts
steoo, line 1: just thinking out loud, is there a way to generate types from the ORM? With graphql and `codegen` you can – it's very handy and keeps things consistent

analytics/src/app/layout.tsx
jonidelv: Good point. In our case, the mismatch is coming from browser extensions that modify the `<body>` tag, not from our app logic. Since we’re not rendering any dynamic content directly in `<body>` and don’t rely on hydration inside it, this should be safe. Going to add a comment to clarify that it’s intentional and only to prevent noise from external factors.

analytics/src/models/Transaction.ts
jonidelv: Yes, but when using server actions, we cannot directly return the model provided by the ORM (Mongoose) due to certain limitations, server actions can only return plain objects. As a result, I need to create a new type and manually transform the data. I'm considering ways to streamline this process, possibly by using Zod for validation and transformation.

analytics/src/app/actions/summary.ts
```diff
@@ -131,19 +125,35 @@ export async function GET(request: Request) {
...
+    const serializedRecentTransactions = recentTransactions.map(transaction => ({
+      _id: transaction._id?.toString() || '',
+      txDate:
+        transaction.txDate instanceof Date ? transaction.txDate.toISOString() : transaction.txDate,
+      sourceTokenId: transaction.sourceTokenId,
+      sourceTokenSymbol: transaction.sourceTokenSymbol,
+      sourceTokenAmount: transaction.sourceTokenAmount,
+      sourceTokenAmountUsd: transaction.sourceTokenAmountUsd,
+      sourceChainUid: transaction.sourceChainUid,
+      sourceChainName: transaction.sourceChainName,
```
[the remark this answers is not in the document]
jonidelv: The API returns the model as a single entity, so I was initially returning it as is without selecting individual properties. However, now that we’ve switched to server actions and need to return plain objects, we have to explicitly deserialize the data and pick only the fields we need. It’s a bit tedious, but it makes it clear exactly what we’re sending and returning.

analytics/README.md
jonidelv: I'm going to rename it to `DOCS.md` because it contains more documentation than anything else.

analytics/src/app/actions/summary.ts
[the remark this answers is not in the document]
Victor-Salomon: I think we could use Zod to `safeParse` or `parse` a schema

app/src/utils/analytics.ts
```diff
@@ -93,18 +89,17 @@ export async function trackTransferMetrics({
     destinationChainNetwork: transferParams.destinationChain.network,
 
     txDate: date,
-    hostedOn: window.location.origin,
+    hostedOn: typeof window !== 'undefined' ? window.location.origin : '',
```
noahjoeris, line 92: Can you explain this to me?

Victor-Salomon (approved):
Solid one. It looks good to me.
Let's see how to integrate this with the widget once we decide on a server solution 🦾.

analytics/src/app/actions/summary.ts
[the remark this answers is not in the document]
jonidelv: I’m exploring how to integrate Zod without having to maintain separate models for Mongoose and Zod. There’s a package that helps bridge the two, and I’m currently looking into it. That said, even with Zod in place, we’d still need to shape the response in this case, since we’re not returning the full Transaction model just a subset of its properties without the Mongoose specific fields.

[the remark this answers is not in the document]
Victor-Salomon: I thought it would be straight forward. Maybe it does not with it in the end

analytics/src/components/pages/DashboardPage.tsx
```diff
@@ -13,17 +14,7 @@ import formatUSD from '@/utils/format-USD'
...
-        headers: {
-          Authorization: process.env.NEXT_PUBLIC_AUTH_TOKEN || '',
-        },
-      })
-      if (!response.ok) {
-        throw new Error(`API error: ${response.status}`)
-      }
-      return await response.json()
-    },
+    queryFn: getSummaryData,
```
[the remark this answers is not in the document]
jonidelv: Reviewed the server actions as you suggested, and it turns out they are more secure, simpler, and perfect for this scenario.

app/src/utils/analytics.ts
jonidelv: I just wanted to be extra cautious about using `window.location.origin` since we’re in a Next environment, where code can run on both the server and the client. While this particular line should always run on the client, I added the `typeof window !== 'undefined` check to be extra safe, just in case it ends up running in a server context or a future test environment where window might not be defined.

analytics/src/app/actions/summary.ts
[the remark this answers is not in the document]
jonidelv: Actually, it is implemented now. 👌🏻

analytics/src/app/actions/summary.ts
```diff
@@ -131,19 +126,22 @@ export async function GET(request: Request) {
     const avgTransactionValue =
       successfulTransactions > 0 ? totalVolumeUsd / successfulTransactions : 0
 
-    const summary = {
+    // Apply the schema to each transaction
+    const serializedRecentTransactions = recentTransactions.map(transaction =>
+      transactionView.parse(transaction),
```
Victor-Salomon, line 131: 🦾

analytics/src/components/pages/DashboardPage.tsx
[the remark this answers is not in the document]
Victor-Salomon: It's a nice combination with React Query. 
It's probably easier to manage than a server component with suspense and loading states and it provides cache directly 💥

steoo (approved):

analytics/src/models/Transaction.ts
jonidelv: It is possible to generate TypeScript types from Mongoose schemas, though not quite as seamlessly or out of the box as with GraphQL Codegen.
When defining the Mongoose schemas, you can define TS interfaces or types and tie them directly to the schema. Example:

```
import { Schema, model, Document } from 'mongoose';

interface IUser extends Document {
  name: string;
  email: string;
  age?: number;
}

const UserSchema = new Schema<IUser>({
  name: { type: String, required: true },
  email: { type: String, required: true },
  age: Number,
});

const UserModel = model<IUser>('User', UserSchema);
```
This way, UserModel is automatically typed, and you get full IntelliSense and type safety when querying or creating users.

PR #392, opened 2025-05-28
fix: ps sdk upgrade with Builder pattern upgrade 

Victor-Salomon:
This PR addresses [the following issue.](https://linear.app/velocitylabs/issue/VEL-413/refactor-max-button)

It upgrades the PS SDK by applying the new standardized Builder pattern (in the final function only). It also includes fixes for breaking changes and updates the Max button according to the linear task.


steoo (approved):

app/src/hooks/useTransferForm.ts
```diff
@@ -230,25 +231,65 @@ const useTransferForm = () => {
...
-        // Parse as number, then format to our display standard, then parse again as number
-        amount: Number(formatAmount(Number(balanceData.formatted), 'Longer')),
-      },
-      { shouldValidate: true },
-    )
-  }, [sourceWallet?.isConnected, sourceTokenAmount?.token, balanceData, setValue])
+    if (sourceChain.network === 'Polkadot') {
+      if (!destinationChain || !destinationWallet?.sender || !sourceWallet?.sender || !sourceToken)
+        return
+
```
noahjoeris, line 247: I don't like nested if's.
Can we do something like:

```
if (sourceChain.network !== 'Polkadot') {
  setValue(
    'sourceTokenAmount',
    {
      token: sourceTokenAmount.token,
      amount: Number(formatAmount(Number(balanceData.formatted), 'Longer')),
    },
    { shouldValidate: true },
  )
  return
}

if (!destinationChain || !destinationWallet?.sender || !sourceWallet?.sender || !sourceToken)
  return

...
```

PR #397, opened 2025-05-30
Feat: Token activity overview

jonidelv:
This PR adds a new page to the Analytics dashboard that displays an overview of Token activity, featuring a sortable table where tokens are organized by volume and transactions.
Additionally, enhance the way it calculates indexes to improve query performance.


analytics/src/app/tokens/page.tsx
steoo, line 1: out of curiosity why are we creating `pages` under `components` instead of creating the pages directly in `page`s files? Having `pages` under `components` feels weird to me

jonidelv: I was initially thinking of keeping all the UI logic under the components folder, but I’m not strongly attached to that approach. I’m also fine with placing the logic directly in each page file within the app directory. What do you think?

analytics/src/app/page.tsx
```diff
@@ -1,5 +1,121 @@
-import DashboardPage from '@/components/pages/DashboardPage'
+'use client'
```
jonidelv, line 1: This is just removing the `DashboardPage` comp and bringing it here so we don't have a separate dir for pages, as Stefano suggested

analytics/src/app/tokens/page.tsx
```diff
@@ -0,0 +1,34 @@
+'use client'
```
jonidelv, line 1: Same here as above

steoo (approved):

PR #399, opened 2025-06-04
Feat: Add volume and transaction count graphs

jonidelv:
Currently, the dashboard home displays statistics based solely on volume. We want to enhance this by including the same graphs that represent transaction count as well. This PR introduces a toggle feature, allowing users to switch between viewing data by volume and by the number of transactions.

Also, include a scenario where we need to update the transaction status to fail if the tracking fails.

### DIFF
--- a/analytics/src/components/TransactionChart.tsx
+++ b/analytics/src/components/TransactionChart.tsx
@@ -15,11 +15,12 @@ import formatUSD from '@/utils/format-USD'
 
 const chartColor = primaryColor
 
-interface TransactionVolumeChartProps {
-  data: { month: string; volumeUsd: number }[]
+interface TransactionChartProps {
+  data: { month: string; volumeUsd: number; count: number }[]
+  type: 'volume' | 'transactions'
 }
 
-const CustomTooltip = ({ active, payload, label }: TooltipProps<number, string>) => {
+const CustomTooltip = ({ active, payload, label, type }: TooltipProps<number, string> & { type: 'volume' | 'transactions' }) => {
   if (!active || !payload || !payload.length) {
     return null
   }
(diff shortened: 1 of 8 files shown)


analytics/src/components/TransactionChart.tsx
```diff
@@ -15,27 +15,37 @@ import formatUSD from '@/utils/format-USD'
 
 const chartColor = primaryColor
 
-interface TransactionVolumeChartProps {
-  data: { month: string; volumeUsd: number }[]
+interface TransactionChartProps {
+  data: { month: string; volumeUsd: number; count: number }[]
+  type: 'volume' | 'transactions'
```
noahjoeris, line 20:
```suggestion
  type: GraphType
```

analytics/src/components/TransactionChart.tsx
```diff
@@ -15,27 +15,37 @@ import formatUSD from '@/utils/format-USD'
...
 }
 
-const CustomTooltip = ({ active, payload, label }: TooltipProps<number, string>) => {
+const CustomTooltip = ({
+  active,
+  payload,
+  label,
+  type,
+}: TooltipProps<number, string> & { type: 'volume' | 'transactions' }) => {
```
noahjoeris, line 28:
```suggestion
}: TooltipProps<number, string> & { type: GraphType }) => {
```

app/src/hooks/useParaspellApi.ts
```diff
@@ -407,6 +420,14 @@ const useParaspellApi = () => {
...
       severity: NotificationSeverity.Error,
     })
+
+    if (txId && environment) {
+      updateTransferMetrics({
+        txHashId: txId,
+        status: TxStatus.Failed,
+        environment: environment,
+      })
+    }
```
noahjoeris, line 430: FYI as the environment we only have Mainnet so far. So it is not really needed at the moment.

noahjoeris (approved):
Looking strong ⭐  I left some comments

Victor-Salomon (approved):
Everything looks good.
I just tried it. 
Common! The graph transition is wiiiiild 🤩
Can wait for your Monday review

PR #402, opened 2025-06-05
Chore: add exception capturing for dry-run validation errors

jonidelv:
This PR introduces a mechanism to identify chains that are failing during DryRun and logs these instances as warnings in Sentry. This will help us start pinpointing these chains and provide a clearer understanding of which ones are failing.

### DIFF
--- a/app/src/hooks/useParaspellApi.ts
+++ b/app/src/hooks/useParaspellApi.ts
@@ -113,6 +113,13 @@ const useParaspellApi = () => {
     const validationResult = await validateTransfer(params)
 
     const defaultDryRunMessage = "Transfer may not succeed. DryRun can't be performed."
+    const dryRunCapturePayload = {
+      extra: {
+        sourceChain: params.sourceChain.uid,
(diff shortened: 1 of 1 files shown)


app/src/hooks/useParaspellApi.ts
```diff
@@ -121,21 +128,31 @@ const useParaspellApi = () => {
             : defaultDryRunMessage,
         severity: NotificationSeverity.Warning,
       })
+      captureException(new Error('DryRun Error: Unsupported'), {
+        ...dryRunCapturePayload,
+        level: 'warning',
```
jonidelv, line 133: just using warning type so we can start idetifying the chains that are failing.

app/src/hooks/useParaspellApi.ts
```diff
@@ -121,21 +128,31 @@ const useParaspellApi = () => {
...
+      })
     }
```
jonidelv, line 135: just using warning type so we can start idetifying the chains that are failing.

noahjoeris (approved):

PR #404, opened 2025-06-09
Feat: Add chains activity stats

jonidelv:
This PR introduces a new page that displays chain activity using a custom Sankey graph to visualize the flow of data across different chains. It also includes a table that presents both inbound and outbound transactions, categorized by volume and count.

Additionally, a new graph has been added to the Token section to help identify outliers, tokens with unusually high transaction counts but low volume, and vice versa


analytics/src/components/ChainSankeyGraph.tsx
```diff
@@ -0,0 +1,695 @@
+import { chainsByUid } from '@velocitylabs-org/turtle-registry'
```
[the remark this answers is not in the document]
NunoAlexandre: This looks look :) I would prefer to either drop the percentage and include the actual number of txs or include both. Just the percentage is already suggested by the nature of the graph and it's not super useful info without objective numbers imo

analytics/src/components/AppLayout.tsx
```diff
@@ -40,30 +42,16 @@ export default function AppLayout({ children }: DashboardLayoutProps) {
...
-      href: '/tokens',
-      active: pathname === '/tokens',
-    },
-  ]
+  const activeRoute = routes.find((route: RouteItem) => route.href === pathname)
 
-  const activeRoute = routes.find(route => route.href === pathname)
+  useEffect(() => {
+    routes.forEach(route => {
+      router.prefetch(route.href)
```
steoo, line 49: is this something we need to do manually in nextjs? for what I've seen on other projects they're prefetched for every link – or am I misunderstanding this?

analytics/src/components/AppLayout.tsx
```diff
@@ -157,20 +146,20 @@ function NavigationMenu({ routes, onNavItemClicked, title = 'Overview' }: Naviga
               prefetch={true}
               className={cn(
                 'group flex w-full cursor-pointer justify-start rounded-lg p-3 text-sm font-medium transition hover:bg-muted hover:text-primary',
-                route.active ? 'bg-muted text-primary' : 'text-muted-foreground',
+                isActiveRoute(route) ? 'bg-muted text-primary' : 'text-muted-foreground',
               )}
-              onClick={onNavItemClicked}
+              onClick={() => onNavItemClicked(isActiveRoute(route))}
```
steoo, line 151: calling this imperatively feels weird, what about the `useMemo` or if really needed the `useEffect` (but it feels like redundant)? if `route` change it should trigger a re-render no?

analytics/src/components/AppLayout.tsx
```diff
@@ -125,20 +113,21 @@ export default function AppLayout({ children }: DashboardLayoutProps) {
...
 interface NavigationMenuProps {
   routes: RouteItem[]
   title?: string
-  onNavItemClicked: () => void
+  onNavItemClicked: (isActive: boolean) => void
+  pathname?: string
 }
 
-function NavigationMenu({ routes, onNavItemClicked, title = 'Overview' }: NavigationMenuProps) {
+function NavigationMenu({
```
steoo, line 123: not a huge fan of having 2 differently named components inside one single file, especially if the file is the `Layout`, I wouldn't think of looking for it in the AppLayout if I didn't know it was there

Victor-Salomon (approved):
LGTM 🦾
Long files to review but preview is once again stunning 🔥. 
(I've left one question about a potential up-coming feature.)

analytics/src/components/AppLayout.tsx
```diff
@@ -125,20 +113,21 @@ export default function AppLayout({ children }: DashboardLayoutProps) {
...
+  routes,
+  onNavItemClicked,
+  title = 'Overview',
+  pathname,
+}: NavigationMenuProps) {
+  const isActiveRoute = (route: RouteItem) => route.href === pathname
```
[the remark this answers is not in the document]
jonidelv: Good catch! Just to clarify: I'm using React Compiler in this analytics project, which automatically adds memo, useMemo, and useCallback optimizations under the hood.

analytics/src/components/AppLayout.tsx
jonidelv: You'd be right that this could cause unnecessary re-renders. However, since I'm using React Compiler on this project, it automatically transforms this into the proper useCallback under the hood, so it stays performant and doesn't trigger unnecessary re-renders.

steoo (approved):

analytics/src/components/AppLayout.tsx
steoo: See your point and that's fair – needs a bit of alignment on style, nothing we can't do. Thank you for the change!

steoo:
Awesome work, lgtm!

PR #411, opened 2025-06-13
Feat: Move State from useState to URL for Shareability

jonidelv:
This PR refactors how component state is managed by introducing [nuqs](https://nuqs.47ng.com/) to store state in the URL rather than using internal useState. This change improves shareability, enabling users to share specific application states via URL.

Changes
	•	Replaced local useState with nuqs for managing state in the URL
	•	Ensured state is synced correctly across navigation and reloads
	•	Performed minor cleanup and improvements in the Select component

### DIFF
--- a/analytics/src/app/transactions/page.tsx
+++ b/analytics/src/app/transactions/page.tsx
@@ -4,12 +4,13 @@ import { Token } from '@velocitylabs-org/turtle-registry'
 import { tokensById, chainsByUid } from '@velocitylabs-org/turtle-registry'
 import { getOriginBadge } from '@velocitylabs-org/turtle-ui'
 import { CheckCircle, X, DollarSign, Ban, CircleHelp } from 'lucide-react'
-import React, { useState } from 'react'
+import { useQueryState, parseAsStringLiteral, parseAsIsoDate } from 'nuqs'
+import React from 'react'
 import { getTransactionsData } from '@/app/actions/transactions'
 import DatePicker from '@/components/DatePicker'
 import ErrorPanel from '@/components/ErrorPanel'
-import MultiSelect from '@/components/MultiSelect'
 import RecentTransactionsTable from '@/components/RecentTransactionsTable'
+import Select from '@/components/Select'
 import SmallStatBox from '@/components/SmallStatBox'
 import { Button } from '@/components/ui/button'
 import { Card, CardContent, CardDescription, CardHeader, CardTitle } from '@/components/ui/card'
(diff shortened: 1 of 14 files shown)


steoo:
Any reason we're using `nuqs` instead of `zustand` like we do in the app or widget? Not a critique, just for my understanding :)

analytics/src/app/transactions/page.tsx
```diff
@@ -91,19 +105,22 @@ export default function TransactionsPage() {
...
+    setFromDate(null)
+    setToDate(null)
   }
 
   if (error && !isLoading) {
     return <ErrorPanel error={error} />
   }
 
+  console.log('sourceChainUid', sourceChainUid)
+  console.log('chainOptions', chainOptions)
```
noahjoeris, line 122:
```suggestion
```

analytics/src/app/transactions/page.tsx
```diff
@@ -19,24 +20,37 @@ import { TxStatus } from '@/models/Transaction'
...
+  const [destinationChainUid, setDestinationChainUid] = useQueryState(
+    'destChain',
+    emptyDefaultString,
+  )
+  const [sourceTokenId, setSourceTokenId] = useQueryState('sourceToken', emptyDefaultString)
+  const [destinationTokenId, setDestinationTokenId] = useQueryState('destToken', emptyDefaultString)
+  const [statusFilterRaw, setStatusFilterRaw] = useQueryState('status', statusFilterParser)
+  const statusFilter = statusFilterRaw === 'all' ? null : (statusFilterRaw as TxStatus | null)
+  const [fromDate, setFromDate] = useQueryState('fromDate', parseAsIsoDate)
+  const [toDate, setToDate] = useQueryState('toDate', parseAsIsoDate)
```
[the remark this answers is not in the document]
jonidelv: Yes, I believe so. Also, nuqs controls each query string parameter separately, I didn't see that they have support for object like states. I prefer keeping things separate for better readability. If they have something in common like form state, then yes I'll use a single state object, but since state updates are now batched, it makes sense to separate them by concerns and contexts.

analytics/src/app/transactions/page.tsx
jonidelv: Good catch

jonidelv:
> Any reason we're using `nuqs` instead of `zustand` like we do in the app or widget? Not a critique, just for my understanding :)

Nuqs is a search params state manager for React, not quite the same as Zustand. We could probably make Zustand work like nuqs, but we'd need to add a lot of work under the hood. Basically, it's like useState but with URL search parameter sync. I feel it's cleaner, plus it's well maintained. Actually,  @Victor-Salomon  recommended using this and I feel it turned out to work quite well. We're adding this so we can share the URL at a specific state of the app :)

Victor-Salomon (approved):
Everything lgtm 🦾

PR #414, opened 2025-06-18
Fix pnpm lock

jonidelv:


NunoAlexandre:
@jonidelv what happened that the pnpm lock got screwed and still we got stuff merged to main?

PR #421, opened 2025-06-19
analytics: Fix db indexes & cleanups

jonidelv:
This PR fixes the database indexes, adds some comments and cleanups

### DIFF
--- a/analytics/src/models/Transaction.ts
+++ b/analytics/src/models/Transaction.ts
@@ -55,7 +55,7 @@ export interface TransactionModel {
 
   txDate: Date
   hostedOn: string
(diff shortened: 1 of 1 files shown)


NunoAlexandre (approved):
The changes are good but ideally we wouldn't need to store info like `isSwap` since it's deductible from the other data. The source of truth should be computable, not stored. #nit

jonidelv:
I was thinking more from a query performance perspective, but that's a valid point. I've already set it up this way, but it can be undone (running a migration). Let me know your thoughts - > if you think this approach is acceptable or if you'd prefer to undo it and go with the computable option

PR #428, opened 2025-07-01
widget: Extend config to filter chains & tokens 

steoo:


widget/src/components/Widget.tsx
```diff
@@ -1,18 +1,28 @@
...
-const Widget = ({ theme }: { theme?: WidgetTheme }) => {
+const Widget = ({
+  theme,
+  allowedChains,
+  allowedTokens,
+}: {
+  theme?: WidgetTheme
+  allowedChains?: Chain['uid'][]
+  allowedTokens?: Token['id'][]
+}) => {
```
NunoAlexandre, line 25: Not a big fan of this model. Proposing the following two options, open to discussion of course.

### Option A
```suggestion
const Widget = ({
  theme,
  registry,
}: {
  theme?: WidgetTheme
  registry?: WidgetRegistry
}) => {
```

and

``` typescript
export interface WidgetRegistry {
    /** A set of supported chain uids based on the Turtle Global Registry. Defaults to the full set if undefined */
    chains?: string[];
    /** A set of supported token ids based on the Turtle Global Registry. Defaults to the full set if undefined */
    tokens?: Token[];
}
```

### Option B

Basically the same but going a little further for a nicer dev ex:

```suggestion
const Widget = ({
  theme,
  filter,
}: {
  theme?: WidgetTheme
  filter?: WidgetFilter
}) => {
```

and

``` typescript
export interface WidgetRegistry {
    /** A set of supported chain uids based on the Turtle Global Registry. Defaults to the full set if undefined */
    chains?: Filter<string[]>;
    /** A set of supported token ids based on the Turtle Global Registry. Defaults to the full set if undefined */
    tokens?: Filter<Token[]>;
}

export type Filter<V> = { type: 'Keep', value: V } | { type: 'Drop', value: V }
```

And then based on the `Filter`.`type`, we filter the chains and the tokens accordingly.

Option B if of course more complex but it grants a really easy way to filter by whitelisting AND blacklisting style, at will. Suppose that some team only wants to drop Ethereum from the list of tokens - instead of having to list N chain ids, they would only have to blacklist ethereum in that scenario.

NunoAlexandre (changes requested):
Besides what I left below, it's also worth checking with Polimec about making Polimec show as the top chain on the list. I am sure they and other teams will want that and it may make sense to think already how to include that in the config model.

widget/src/components/Widget.tsx
steoo: I'd go for Option A if you feel it's more compact because Option B needs a little more thinking and could be a next iteration of it, not in love with the `type` approach, like if it was a command – if we were to take this route I'd rather do something like

```
include: []
exclude: []
```

Regarding Option A, is `registry` a definition that's common in the ecosystem? I used `allowedChains` and `allowedTokens` for how explicit it is.

steoo:
> Besides what I left below, it's also worth checking with Polimec about making Polimec show as the top chain on the list. I am sure they and other teams will want that and it may make sense to think already how to include that in the config model.

The order was something that was also mentioned now that I think about it yes

widget/src/components/Widget.tsx
NunoAlexandre: > like if it was a command

I like that ahah but sure. 


> not in love with the type approach

Why? I think it's one of the few nice type features that JS offers that comes close to union types. 

```
include: []
exclude: []
```

Unlike the type model I proposed, this is not mutually exclusive and it would let users specify both if they choose to and require a runtime check to invalidate that scenario. No bueno.


Let's go with Option A then 👍

steoo: You can make them mutually exclusive tho! The main difference is that basically you either take it from `include` or `exclude` instead of having to check for `type` which to me feels like an unnecessary step – does it make sense? 

Let's keep it simple with Option A

steoo:
> > > Besides what I left below, it's also worth checking with Polimec about making Polimec show as the top chain on the list. I am sure they and other teams will want that and it may make sense to think already how to include that in the config model.
> > 
> > 
> > The order was something that was also mentioned now that I think about it yes
> 
> Yeah let's implement it. The caveat is that we shouldn't mix the Option A / filtering and setting a weight for sorting. For example, a team might want to embed the widget and show all chains and all tokens but have chains `X, Y, Z` and token `A, B, C` on top. So it would be nice for users to say : X -> weight 3, Y, -> weight 2, Z -> weight 1, and so on, **without** it meaning that only chains X, Y and Z should be shown.

Could be a sorting option easily – but yeah let's not mix it

widget/src/components/Widget.tsx
NunoAlexandre: > You can make them mutually exclusive tho!

How? 👀

widget/README.md
```diff
@@ -120,6 +120,86 @@ You can pass a theme prop to `<Widget />` to override default styles like colors
 
 > ⚠️ Note: The Turtle Widget bundles its own Tailwind CSS, but it's not sandboxed. Global styles from your app (e.g., `.your-wrapper p { ... }`) may still affect it. This allows for advanced customization, but be cautious with _aggressive_ global CSS.
 
+## ⛓️ Chain and Token Configuration
+
+You can also customize the Chains and Tokens you want to show in your widget instance – it works by passing the variable registry
```
NunoAlexandre, line 125:
```suggestion
You can also customize the Chains and Tokens you want to show in your widget instance – it works by passing the `registry` configuration. It defaults to using all chains & tokens supported by the [Turtle App](app.turtle.cool) otherwise.
```

widget/README.md
```diff
@@ -120,6 +120,86 @@ You can pass a theme prop to `<Widget />` to override default styles like colors
...
+
+```
+registry = {
+  chains: [],
+  tokens: []
+}
+```
```
NunoAlexandre, line 132:
```suggestion

For example:

\```
const registry = {
  chains: ['polkadot', 'hydration'],
  tokens: ['dot', 'usdc', 'usdt'],
}
\```
```


(We have to drop the '\')

widget/README.md
```diff
@@ -120,6 +120,86 @@ You can pass a theme prop to `<Widget />` to override default styles like colors
...
+astar // Astar
+mythos // Mythos
+```
+
+#### Tokens ids
+
+```
+eth // Ethereum
+usdc.e // USD Coin (bridged)
+dai.e // DAI (bridged)
```
NunoAlexandre, line 162: It would be nice to offer a very simple page that dynamically displays these ids instead of having us manually copy-paste them here. For example, it could be accessible at `registry.turtle.cool` and basically be a json prettified version of the registry or so. Let's add that to the backlog.

widget/src/components/Transfer.tsx
```diff
@@ -121,6 +122,8 @@ const Transfer: FC = () => {
     applyTransferableBalance,
   } = useTransferForm()
 
+  const { allowedChains, allowedTokens } = use(ConfigContext)
```
NunoAlexandre, line 125: I think there's no need for `allowed` in these names. First, it's not about being "allowed", it's just filtered. so either rename it to `filteredChains` / `filteredTokens` or just `chains` and `tokens`.

widget/src/providers/ConfigProviders.tsx
```diff
@@ -0,0 +1,31 @@
...
+  tokens: Token['id'][]
+}
+
+export const ConfigContext = createContext<{
+  allowedChains: Chain['uid'][]
+  allowedTokens: Token['id'][]
+}>({
+  allowedChains: [],
+  allowedTokens: [],
+})
```
NunoAlexandre, line 15:
```suggestion
export const ConfigContext = createContext<{
  chains: Chain['uid'][]
  tokens: Token['id'][]
}>({
  chains: [],
  tokens: [],
})
```

Is there no way to not have to repeat the fields?

widget/src/utils/routes.ts
```diff
@@ -20,36 +20,19 @@ const deduplicate = <T extends { uid?: string; id?: string }>(items: T[]): T[] =
...
-        route.from === sourceChain.uid && route.tokens.includes(token.id) && route.to === c.uid,
-    ),
-  )
-}
-
 /** Filters all tokens by selected source chain and available routes */
 export const getTransferTokens = (
   sourceChain: Chain | null,
   destinationChain: Chain | null,
+  allowedTokens?: Token['id'][] | undefined,
```
NunoAlexandre, line 27:
```suggestion
  filteredTokens?: Token['id'][] | undefined,
```

In this context it makes sense to have the `filtered` prefix to give context to the logic.

NunoAlexandre (commented):
Getting there 👌

widget/src/providers/ConfigProviders.tsx
steoo: what you mean? 🤔

widget/README.md
steoo: Like a small dashboard with our registry visualised? That would be cool – maybe there's something similar already existing in the Analytics?

widget/src/utils/routes.ts
NunoAlexandre: ah I see. No, let it be, keep it as is.

NunoAlexandre (approved):
👌

PR #430, opened 2025-07-03
Feat: Add swaps page in analytics dashboard

jonidelv:
This PR introduces a dedicated page for swaps, displaying recent swaps, swap rates, and metrics for inbound and outbound token activity.

### DIFF
--- a/analytics/src/app/actions/swaps.ts
+++ b/analytics/src/app/actions/swaps.ts
@@ -1,0 +1,166 @@
+'use server'
+import { defaultTransactionLimit, swapsStartingDate } from '@/constants'
+import swapView from '@/models/swap-view'
+import Transaction from '@/models/Transaction'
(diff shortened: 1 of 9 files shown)


NunoAlexandre:
@jonidelv can you:

- move the Swaps tab to below "Transactions" in the left menu
- Show more swaps on 'Recent swaps' (now it's 10, make it 50) 

Beside that, it feels weird to me that 'Swaps' is a separate thing from 'Transactions' but it's not a blocker for now 👍

jonidelv:
@NunoAlexandre Sounds good! about showing 50 swaps -> I was planning to wait until we have pagination in place, especially since we currently only have 11 swaps. But let’s do this: I’ll update it to show the last 20 for now, and once pagination is in, we’ll be able to paginate through all swaps and transactions smoothly and pagination is coming pretty soon.

PR #431, opened 2025-07-04
Impr: Replace AreaChart with BarChart

jonidelv:
This PR changes the type of graph displayed on the dashboard for cumulative volume and the number of transactions. We are switching from an area graph to a more common and understandable bar graph. Additionally, we are excluding the current day and month from the data to prevent the graph from showing a downward trend when there isn't enough information available.


NunoAlexandre (approved):
👌

PR #436, opened 2025-07-09
Fix Widget Registry entry points 

steoo:


widget/src/main.tsx
```diff
@@ -7,7 +7,12 @@ import './index.css'
 createRoot(document.getElementById('root')!).render(
   <StrictMode>
     <div className="flex h-screen w-screen items-center justify-center">
-      <Widget />
+      <Widget
```
Victor-Salomon, line 10: Should we remove this? :)

Victor-Salomon (approved):
LGTM

widget/.env.local.example
```diff
@@ -0,0 +1,10 @@
+
+# Provide your WalletConnect project ID to allow access to WalletConnect Cloud.
+# VITE_PUBLIC_SENTRY_DSN=
+VITE_PUBLIC_WALLETCONNECT_PROJECT_ID=
+VITE_PUBLIC_SUBSCAN_KEY=
+VITE_PUBLIC_ALCHEMY_KEY=
+VITE_PUBLIC_DWELLIR_KEY=
+VITE_PUBLIC_ENVIRONMENT= # eg: 'mainnet'
+VITE_PUBLIC_FIREBASE_PROJECT_ID= # eg: 'app-turtle-cool'
+VITE_PUBLIC_OC_API_KEY_READ_WRITE=
```
jonidelv, line 10: Let's add here the two env variables for analytics (pasted them in Slack)

widget/src/main.tsx
jonidelv: Maybe we could just default to our registry for testing locally the widget, so we can see all the options

PR #468, opened 2025-07-25
fix: tracking the execute function correctly

noahjoeris:
This PR fixes the execute tracking for swap + send flows.


Victor-Salomon (approved):
✅

PR #470, opened 2025-07-28
Refactor: Improve UI details in Analytics

jonidelv:
This PR improves several UI details that I set aside until the end of the project, primarily on the transaction detail page. These enhancements will allow for better tracking of transaction failures. Additionally, I've made some fixes related to database querying and the main graph.
Also, added a new option called "flattened" that will flatten the curve and remove outliers from the main graph (filtering transactions > 50k)


analytics/src/components/ui/checkbox.tsx
steoo, line 1: is this a copy paste of what we have in App/Widget? If so, would you mind moving it to the UI package, import it from there, and delete the original one?

NunoAlexandre:
> Also, added a new option called "flattened" that will flatten the curve and remove outliers from the main graph (filtering transactions > 50k)

Can you rename it to "Outliers" and make it enabled by default. So we if we want to see it without outliners, we check the box out.

steoo (approved):

PR #471, opened 2025-07-28
Chainflip/make app sdk agnostic

Victor-Salomon:


packages/registry/src/mainnet/routes.ts
```diff
@@ -28,8 +46,8 @@ export const routes: Route[] = [
       EthereumTokens.ETH.id,
       EthereumTokens.WETH.id,
       EthereumTokens.WBTC.id,
-      EthereumTokens.USDC.id,
-      EthereumTokens.USDT.id,
+      // EthereumTokens.USDC.id, // now Supported by Chainflip
+      // EthereumTokens.USDT.id, // now Supported by Chainflip
```
NunoAlexandre, line 50:
```suggestion
```

packages/registry/src/mainnet/chainflipPairs.ts
```diff
@@ -0,0 +1,61 @@
...
+import { EthereumTokens, PolkadotTokens } from "./tokens";
+
+export type ChainflipSwapPair = [Token, Token];
+export type ChainflipSwapRoute = {
+  from: Chain;
+  to: Chain;
+  pairs: ChainflipSwapPair[];
+};
+
+export const chainflipSwapPairs: ChainflipSwapRoute[] = [
```
NunoAlexandre, line 12: the const is called `SwapPairs` while the type is `SwapRoute`. I would just call it `Route`. So `chainFlipRoutes` and `ChainFlipRoute`, respectively.

packages/registry/src/mainnet/chainflipPairs.ts
```diff
@@ -0,0 +1,61 @@
...
+      [EthereumTokens.ETH, PolkadotTokens.DOT],
+      [EthereumTokens.ETH, PolkadotTokens.USDC],
+      [EthereumTokens.ETH, PolkadotTokens.USDT],
+      [EthereumTokens.USDC, PolkadotTokens.DOT],
+      [EthereumTokens.USDC, PolkadotTokens.USDC],
+      [EthereumTokens.USDC, PolkadotTokens.USDT],
+      [EthereumTokens.USDT, PolkadotTokens.DOT],
+      [EthereumTokens.USDT, PolkadotTokens.USDC],
+      [EthereumTokens.USDT, PolkadotTokens.USDT],
+    ],
```
NunoAlexandre, line 26: We could make these route entries a map to avoid duplication and reduce search complexity. In pseudo-code:

```suggestion
    pairs: {
      'EthereumTokens.ETH': [PolkadotTokens.DOT, PolkadotTokens.USDC, PolkadotTokens.USD],
      'EthereumTokens.USDC':  [PolkadotTokens.DOT, PolkadotTokens.USDC, PolkadotTokens.USDT, ,
      'EthereumTokens.USDT': [PolkadotTokens.DOT, PolkadotTokens.USDC, PolkadotTokens.USDT],
    }
```

NunoAlexandre (commented):
The changes looks fair and sensible overall, so I am leaving some other comments / questions:

- What is the purpose of the 'standard' field in `token.origin`? What will that piece of information be useful for?
- We need to backwards compatible and thus decode stored data and show the origin badge for ongoing / done txs.

packages/registry/src/mainnet/chainflipPairs.ts
Victor-Salomon: Totally, it's part of my to-do to clean this draft. 
Thanks for notifying 🦾

Victor-Salomon:
> The changes looks fair and sensible overall, so I am leaving some other comments / questions:
> 
> * What is the purpose of the 'standard' field in `token.origin`? What will that piece of information be useful for?
> * We need to backwards compatible and thus decode stored data and show the origin badge for ongoing / done txs.

@NunoAlexandre Thanks for the feedback. I agree, this is a sensitive topic.

Regarding the `standard` field and backward compatibility: for now, both tokens (including their id) and badges behave exactly as they were. The `standard` field allows us to use Ethereum tokens with Chainflip without using Native Eth tokens marked as Snowbridge-wrapped by default.

Since Snowbridge is currently our only bridge for wrapped tokens, the badge logic stays backward-compatible regardless of the origin value we set.

If we want to support a new bridge with wrapped assets, we should be able to dynamically generate a Snowbridge-specific token list and implement it in our `routes` instead of the Eth tokens (Native or ERC20 we now use with Chainflip). This is why I call this approach more future-proof than agnostic.

app/src/__tests__/registry.test.ts
```diff
@@ -13,8 +13,9 @@ describe('Routes', () => {
   })
 
   it('should not have any duplicate routes', () => {
-    const routeSet = new Set(routes.map(r => `${r.from}-${r.to}`))
-    expect(routeSet.size).toBe(routes.length)
+    const filteredRoutes = routes.filter(r => r.sdk !== 'ChainflipApi')
+    const routeSet = new Set(filteredRoutes.map(r => `${r.from}-${r.to}`))
+    expect(routeSet.size).toBe(filteredRoutes.length)
```
noahjoeris, line 18: As we saw we had a duplicate, we definitely need this for chainflip as well! 😉

packages/registry/src/mainnet/tokens.ts
```diff
@@ -26,12 +26,34 @@ import wethLogo from "@velocitylabs-org/turtle-assets/logos/weth.svg";
...
+            ? tokenKey.toLowerCase()
+            : `${tokenKey.toLowerCase()}.${suffix}`,
+        ...tokenValue,
+        origin: ethereumOrigin(tokenKey === "ETH" ? "Native" : "ERC20"),
+      } as Token;
+      return acc;
+    },
+    {} as { [K in keyof T]: Token },
+  );
+};
```
noahjoeris, line 53: Please add a doc comment

app/src/utils/chainflip.ts
```diff
@@ -0,0 +1,57 @@
...
+  if (!sourceChain) return []
+
+  const tokensSet = new Set<Token>()
+
+  chainflipRoutes
+    .filter(route => route.from.chainId === sourceChain.chainId)
+    .map(route => route.pairs.map(([token, _]) => tokensSet.add(token)))
+
+  return Array.from(tokensSet)
+}
```
noahjoeris, line 19: You don't need to use a `Set` here. You can directly use an `Array`.

app/src/utils/chainflip.ts
```diff
@@ -0,0 +1,57 @@
+import { Chain, chainflipRoutes, Token } from '@velocitylabs-org/turtle-registry'
+
+/** returns all Chainflip allowed source chains for a swap. */
+export const getChainflipSwapSourceChains = () => {
```
noahjoeris, line 4: Could you add the return type to all the functions inside chainflip.ts? 🙏

noahjoeris (changes requested):
Awesome work Victor. I left some changes 🔥

packages/registry/src/mainnet/tokens.ts
NunoAlexandre: yeah this code is rough waters

app/src/utils/chainflip.ts
noahjoeris: If the intention was to not have duplicates it doesn't work when you use objects and not IDs

app/src/utils/chainflip.ts
Victor-Salomon: I forgot to add them! thanks.

packages/registry/src/types/token.ts
```diff
@@ -22,8 +22,10 @@ export interface Token {
 
 // The supported bridges. Each bridge will have it's own wrapped version of a token
 export type Bridge = "Snowbridge" | "Other";
+export type TokenStandard = "Native" | "ERC20"; // keep it future proof - Ex: "SPL" Solana Standard
 
 // The origin of a token
 export type Origin =
+  | { type: "Ethereum"; standard: TokenStandard }
   | { type: "Ethereum"; bridge: Bridge }
```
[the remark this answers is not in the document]
Victor-Salomon: sure, I'll add some documentation

packages/registry/src/mainnet/tokens.ts
Victor-Salomon: hehe sure, I'll coment on all of this.

PR #474, opened 2025-08-01
Fix: Add prettier-plugin-tailwindcss to devDependencies

jonidelv:


jonidelv:
Prettier was not working for me because this necessary package was missing. This solution resolves the issue. While I agree about moving away from Prettier, I believe this fix will help in the short term, especially since I'm experiencing some deployment issues with that pull request (removing prettier and eslint) and it may take more time to address those.

PR #476, opened 2025-08-01
Fix: landing volume ui

jonidelv:
Currently, there is a small flicker where the volume starts at 0 and then changes to the actual volume. This issue arises from the way the fetch function is executed, which causes a problem when retrieving data from the server. The solution to this is to ensure that the initial volume is fetched from the server. This way, when the app renders, it already has the correct volume, and updates will occur in real time if necessary.

### DIFF
--- a/site/src/components/AnalyticData.tsx
+++ b/site/src/components/AnalyticData.tsx
@@ -12,7 +12,7 @@ const format = {
   trailingZeroDisplay: 'stripIfInteger',
 } as const
 
-const defaultVolume = 10123001
+const lastKnownVolume = 10123001
 
 interface AnalyticDataProps {
   initialVolume: number | undefined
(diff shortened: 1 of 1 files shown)


NunoAlexandre:
@jonidelv it does 👍 

but when I open the page for the first time it still starts with volume '0'. We can hardcode a default last known volume to date  and have it start there until it loads the most recent amount.

jonidelv:
Could you please clear your cache and try again? This issue should be resolved

NunoAlexandre:
> Could you please clear your cache and try again? This issue should be resolved

Users won't do that so that's not the solution

NunoAlexandre:
> Visit Preview

oh I see what you mean. you mean that probably I had an older version of the site. gotcha.

NunoAlexandre (approved):
🚀  Let's go!

PR #478, opened 2025-08-04
chainflip: generate quote and implement swap output amount

Victor-Salomon:
This PR: 

- Implement chainflip swap routes to the chain and token list.

- Implement the chainflip SDK to provide the output amount:
1. Store chainflip client
2. Cache transfer quote
3. Handle minimum amount error

(Quote is a work in progress and might be updated. Don't care about logs and potential comments)


jonidelv (approved):
lgtm 👏

PR #483, opened 2025-08-07
Refactor: Enhance XCM fees

jonidelv:
This PR updates the useFee hook, specifically regarding how we fetch the fees from Paraspell. It now uses their new method directly from the builder for a more accurate (dryrun) result.

### DIFF
--- a/app/src/hooks/useFees.ts
+++ b/app/src/hooks/useFees.ts
@@ -12,7 +12,7 @@ import { Direction, resolveDirection } from '@/services/transfer'
 import { getNativeToken, getParaSpellNode, isChainSupportingToken } from '@/utils/paraspellTransfer'
 import { resolveSdk } from '@/utils/routes'
 import { getFeeEstimate } from '@/utils/snowbridge'
-import { convertAmount, toHuman } from '@/utils/transfer'
+import { safeConvertAmount, toHuman } from '@/utils/transfer'
 import useBalance from './useBalance'
 import useSnowbridgeContext from './useSnowbridgeContext'
 
@@ -99,7 +99,7 @@ const useFees = (params: UseFeesParams) => {
             sourceChain: sourceChain,
             destinationChain: destinationChain,
             sourceToken,
-            sourceAmount: convertAmount(sourceTokenAmount, sourceToken),
+            sourceAmount: safeConvertAmount(sourceTokenAmount, sourceToken)!,
             sender,
             recipient: recipientAddress,
           })


app/src/hooks/useFees.ts
```diff
@@ -82,24 +95,29 @@ const useFees = (
...
-            destination: destinationChainNode,
-            currency: { ...currency, amount: BigInt(10 ** token.decimals).toString() }, // hardcoded amount because the fee is usually independent of the amount
-            account: getPlaceholderAddress(sourceChain.supportedAddressTypes[0]), // hardcode sender address because the fee is usually independent of the sender
-            accountDestination: getPlaceholderAddress(destinationChain.supportedAddressTypes[0]), // hardcode recipient address because the fee is usually independent of the recipient
-            api: sourceChain.rpcConnection,
+          const originXcmFee = await xcmTransferBuilderManager.getOriginXcmFee({
+            sourceChain: sourceChain,
+            destinationChain: destinationChain,
+            sourceToken,
+            sourceAmount: convertAmount(sourceTokenAmount, sourceToken),
```
jonidelv, line 102: Convert it to BigInt

app/src/hooks/useFees.ts
```diff
@@ -82,24 +95,29 @@ const useFees = (
...
+            recipient: recipientAddress,
           })
 
+          const {
+            currency: feeCurrency,
+            fee: feeAmount,
+            sufficient: sufficientForXCM,
+          } = originXcmFee
+
+          const feeToken = PolkadotTokens[feeCurrency! as keyof typeof PolkadotTokens]
```
jonidelv, line 113: `feeCurrency` is just the symbol, here we get the fees token

app/src/hooks/useFees.ts
```diff
@@ -126,7 +144,13 @@ const useFees = (
...
 
         case 'SnowbridgeApi': {
-          if (!sourceChain || !senderAddress || !destinationChain || !amount || !recipientAddress) {
+          if (
+            !sourceChain ||
+            !senderAddress ||
+            !destinationChain ||
+            !sourceTokenAmount ||
+            !recipientAddress
+          ) {
```
jonidelv, line 153: Just different formatting, since I renamed some parameters in this hook

app/src/hooks/useFees.ts
```diff
@@ -188,11 +212,11 @@ const useFees = (
...
-      // addNotification({
-      //   severity: NotificationSeverity.Error,
-      //   message: 'Failed to fetch the fees. Please try again later.',
-      //   dismissible: true,
-      // })
+      addNotification({
+        severity: NotificationSeverity.Error,
+        message: 'Failed to fetch the fees. Please try again later.',
+        dismissible: true,
+      })
```
jonidelv, line 219: I believe this notification is beneficial when fees fail for any reason. I'm not sure why it was commented out.

app/src/services/paraspell/xcmTransferBuilder.ts
```diff
@@ -114,6 +118,37 @@ class XcmTransferBuilderManager {
     }
   }
 
+  // Origin and destination fees
+  async getXcmFee(
```
jonidelv, line 122: We are not using this one for now, but it will be beneficial for the future

steoo (approved):

app/src/hooks/useFees.ts
noahjoeris: multilocation would be safer to use than symbol

app/src/hooks/useFees.ts
```diff
@@ -82,24 +95,29 @@ const useFees = (
...
+            fee: feeAmount,
+            sufficient: sufficientForXCM,
+          } = originXcmFee
+
+          const feeToken = PolkadotTokens[feeCurrency! as keyof typeof PolkadotTokens]
           const feeTokenInDollars = (await getCachedTokenPrice(feeToken))?.usd ?? 0
-          const fee = info.xcmFee
           setFees({
-            amount: fee,
+            amount: feeAmount!,
```
noahjoeris, line 116: why do we need the '!' ?

noahjoeris (approved):
Nice one mate 🔥

Victor-Salomon (approved):
LGTM.
I can already identify some conflits with my work, but once fixed, I will benefit from the improvments as well 🦾

Nice one

app/src/hooks/useFees.ts
jonidelv: Unfortunatly `getOriginXcmFee` just returns the symbol

app/src/hooks/useFees.ts
jonidelv: TS indicated that the feeAmount could be `bigint | undefined`, but it was actually always defined. I could implement a check to throw an error if it is undefined.

PR #495, opened 2025-08-15
Refactor: useFees Hook to Support Fee Arrays

jonidelv:
This PR updates our fee calculation approach.
Previously, we had separate `fee` and `bridgingFees` values. Now, we’re using Paraspell’s more detailed fees function, which provides origin fees, destination fees, and hop fees.

To support this, the fee format has been refactored into a single array of fees, resulting in changes across the app.
A migration has also been added for older transactions stored in local storage (Zustand storage).


app/src/components/Transfer.tsx
```diff
@@ -114,8 +113,7 @@ export default function Transfer() {
     balanceData,
     fetchBalance,
     isLoadingOutputAmount,
-    exceedsTransferableBalance,
-    applyTransferableBalance,
+    maxButtonLoading,
```
jonidelv, line 116: Added a loading phase to the max button, it sometimes took a few seconds, so the loading indicator was a nice addition

app/src/hooks/useFees.ts
```diff
@@ -65,129 +78,184 @@ const useFees = (params: UseFeesParams) => {
...
             recipientAddress,
             sourceTokenAmount,
           )
-          if (!fee) {
+          if (!snowbridgeFees || snowbridgeFees.origin !== 'Ethereum') {
+            throw new Error('getFeeEstimate could not run successfully')
+          }
+
+          // Check if this is an insufficient funds error where the amount is not enough to cover both the transfer fees and the requested amount
+          if ('error' in snowbridgeFees && snowbridgeFees.error === 'INSUFFICIENT_FUNDS') {
```
jonidelv, line 216: This error happens when the amount is not enough to cover both the transfer fees and the requested amount

app/src/hooks/useFees.ts
```diff
@@ -210,48 +290,82 @@ const useFees = (params: UseFeesParams) => {
...
+    return 0
+  }
+}
 
-  if (!response.ok) {
-    const { error } = await response.json()
-    throw new Error(error || `Failed to fetch bridging fee`)
+function getTokenFromSymbol(symbol: string): Token {
+  const symbolUpper = symbol.toUpperCase()
+  const tokensBySymbol = { ...EthereumTokens, ...PolkadotTokens }
```
jonidelv, line 326: Paraspell provides only the symbol, but since these symbols are repeated and Paraspell is used within the Polkadot ecosystem, we prioritize Polkadot tokens over Ethereum tokens: `{ ...EthereumTokens, ...PolkadotTokens }`.

app/src/utils/transfer.ts
```diff
@@ -51,15 +51,6 @@ export function feeToHuman(fees: AmountInfo): string {
   return toHuman(fees.amount, fees.token).toFixed(10)
 }
 
-export const getTotalFees = (fees: AmountInfo, bridgingFee?: AmountInfo | null) => {
```
jonidelv, line 54: We were not using this anywhere, and with the fees changed, it is safer to remove it.

packages/assets/logos/ausdt.svg
jonidelv, line 1: I know this isn't directly related to the task, but I think it's an easy win. The old icon had too much white space, making it difficult to notice the "t" in USDT. The new icon fills the entire space and looks much better.

Image

app/src/models/transfer.ts
```diff
@@ -154,3 +193,19 @@ export type OnChainBaseEvents = {
   isExtrinsicSuccess?: boolean
   isExecuteAttemptCompleted?: boolean
 }
+
+type FeeDetailType = 'Execution fees' | 'Destination fees' | 'Bridging fees' | 'Swap fees'
+
+export type FeeSufficiency = 'sufficient' | 'insufficient' | 'undetermined'
```
jonidelv, line 199: `sufficient` may be undefined when chains lack dry-run support, which prevents fee/success validation. 
Present: reliable sufficiency indicator 
Missing: defaults to true, but the user proceeds at their own risk (we show a warning)

PR #486, opened 2025-08-08
redesign: New TxSummary component

NunoAlexandre:
`WIP` 🚧 

Closes https://linear.app/velocitylabs/issue/VEL-471/implement-new-txsummary

### Preview
https://github.com/user-attachments/assets/d1ffdf26-5eec-41fb-b863-23b1ef939f2c

### To Do

- [ ] Ideally, we just update the receiving amount (token and in dollars) and the fee when the from token amount is changed, instead of reloading the whole thing
- [ ] Implement Error screens
- [ ] Plug in @Victor-Salomon / @jonidelv's work with the new fee's breakdown (mock in place for now)

PR #501, opened 2025-08-22
Refactor: New Tx summary

jonidelv:
This PR introduces a new Transaction Summary UI with the updated fees format.
[Closes #VEL-471](https://linear.app/velocitylabs/issue/VEL-471/implement-new-txsummary)

This work continues from the previous PR https://github.com/velocitylabs-org/turtle/pull/486.


Victor-Salomon:
The overall design is stunning 😎

jonidelv:
Hey Victor, 
I'm trying to reproduce the issue, but I'm not able to. It seems that something happened, and the fees have failed on your end, which is why the icon for the fees breakdown is not appearing. I can modify the UI to remove the word "fees" altogether if the fees fail. However, I believe we should still display the receiving amount, even if the fees don’t appear. Additionally, I could add a warning to indicate that the fees have failed.

noahjoeris:
Got the same issues on the preview as Victor described

PR #503, opened 2025-08-26
Fix: Handle Moonbeam ERC-20 prefixes in token symbol resolution

jonidelv:
This PR strip the `xc` prefix and map to the corresponding base token (e.g., xcDOT → DOT).

### Related Issue
Close [VEL-495](https://linear.app/velocitylabs/issue/VEL-495/map-xcdot-fees-to-dot-when-transferring-hydration-%E2%86%92-moonbeam)


noahjoeris:
Why don't we map by multilocatoin instead of symbol?

PR #509, opened 2025-09-02
Fix: Integrate 'Dazzed' font within widget

jonidelv:
This PR implements the same font in the Widget as used in the App. This enhances consistency across turtle.


steoo:
it was removed on purpose :) I reckon it's because this way we can use a neutral font-system that works for every potential customer

@Victor-Salomon do you remember why we did it?

jonidelv:
Oh I see, ok let me close the PR

PR #511, opened 2025-09-03
Chainflip swaps

Victor-Salomon:


app/src/components/ongoing/OngoingTransferDialog.tsx
```diff
@@ -37,6 +39,13 @@ export default function OngoingTransferDialog({ transfer, status }: OngoingTrans
     [transfer.status],
   )
 
+  const isChainflipCheck = isChainflipSwap(
+    transfer.sourceChain,
+    transfer.destChain,
+    transfer.sourceToken,
+    transfer.destinationToken,
+  )
+
```
NunoAlexandre, line 48: we are using two terminologies here: `isCheck` and `isSwap`. Let's stick with one. And ideally, we don't need to pass all of these fields because once Snowbridge supports Bridge&Swap, we won't be able to tell if it's a chainFlip swap or a Snowbridge one. We should have the `route` value here and just this would become `isChainFlipTx = route.sdk === 'ChainFlip'`

app/src/components/ChainflipRefund.tsx
```diff
@@ -0,0 +1,59 @@
+import ExclamationIcon from '@velocitylabs-org/turtle-assets/icons/exclamation-circle.svg'
+import InfoIcon from '@velocitylabs-org/turtle-assets/icons/info.svg'
+import type { Chain, Token } from '@velocitylabs-org/turtle-registry'
+import { cn } from '@velocitylabs-org/turtle-ui'
+import Image from 'next/image'
+import { ChainflipRefundStatus, isChainflipSwap } from '@/utils/chainflip'
+
+interface ChainflipRefundProps {
+  isChainflipCheck?: boolean | null
```
NunoAlexandre, line 9: For me, this doesn't make sense. If `isChainflipCheck`, the component should't be rendering at all. So at the call site, it would be:

```
{ isChainFlipTx && ChainflipRefund<...> }
```

app/src/components/Transfer.tsx
```diff
@@ -126,6 +129,29 @@ export default function Transfer() {
...
+      // Transfer is not a Chainflip swap
+      !isChainflipSwap ||
+      // if Chainflip swap is loading or if swap quote is avaialble & valid
+      (isChainflipSwap && isLoadingChainflipQuote) ||
+      (isChainflipSwap && !isChainflipQuoteError && !!chainflipQuote)
+    )
+      return true
+
+    // If conditions are not met, the swap is not allowed
+    return false
```
NunoAlexandre, line 152:
```suggestion
    return (
      // Transfer is not a Chainflip swap
      !isChainflipSwap ||
      // if Chainflip swap is loading or if swap quote is avaialble & valid
      (isChainflipSwap && isLoadingChainflipQuote) ||
      (isChainflipSwap && !isChainflipQuoteError && !!chainflipQuote)
    )
```

app/src/components/Transfer.tsx
```diff
@@ -427,15 +479,31 @@ export default function Transfer() {
...
 
+      {/* Chainflip quote error */}
+      <AnimatePresence>
+        {isChainflipQuoteError && !requiresErc20SpendApproval && (
+          <motion.div className="flex items-center gap-1 self-center pt-1" {...approvalAnimationProps}>
+            <ActionBanner
+              disabled={false}
+              header="Can't swap this pair for now."
+              text={`${chainflipQuoteError?.message} Please try a different pair.`}
+              image={<Image src="/wip.png" alt="Work in progress" width={64} height={64} />}
```
NunoAlexandre, line 490: @Victor-Salomon @raissapax as a user, I would like to have some context here.

NunoAlexandre (commented):
A start, will continue 👍

app/src/components/Transfer.tsx
Victor-Salomon: @NunoAlexandre, I'm not sure what to add on top of the `quote error message` provided by Chainflip + `Please try a different pair.`

image

app/src/components/completed/TransactionDialog.tsx
```diff
@@ -181,6 +184,13 @@ export default function TransactionDialog({ tx }: TransactionDialogProps) {
                   />
                 ))}
 
+              <ChainflipRefund
+                isChainflipCheck={isChainflipCheck}
+                swapCompleted={tx.result === TxStatus.Succeeded}
+                swapRefundError={tx.errors?.[0]}
+                className="pt-5 pb-2"
```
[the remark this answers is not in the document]
Victor-Salomon: Do we display multiple different banners in the Ongoing/Completed Transfers?

app/src/components/Transfer.tsx
NunoAlexandre: @Victor-Salomon sorry, I missed the first line of the explanation, this is fine 🙏

app/src/components/completed/TransactionDialog.tsx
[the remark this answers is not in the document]
NunoAlexandre: No, but we do in the TxSummary, my comment was misplaced. If it's good, it's good 👍

app/src/hooks/useChainflipTracker.ts
```diff
@@ -0,0 +1,160 @@
...
+  addNotification: (notification: Omit<Notification, 'id'>) => void,
+): Promise<void> => {
+  const { sourceToken, sourceChain, destChain, destinationToken, id, uniqueTrackingId, recipient } = swap
+  if (!uniqueTrackingId) throw new Error(`Chainflip swap ${id} does not have a unique tracking id`)
+  const status = await getSwapStatus(uniqueTrackingId)
+  const { state } = status
+  const formattedRecipient = truncateAddress(recipient)
+  switch (state) {
+    case 'RECEIVING': {
+      const newStatus = `${sourceToken.symbol} funds deposited`
```
NunoAlexandre, line 28:
```suggestion
      const newStatus = `${sourceToken.symbol} deposited`
```

I don't think we need "funds" here, as we don't do either in the other messages.

app/src/hooks/useChainflipTracker.ts
```diff
@@ -0,0 +1,160 @@
...
+  const isPollingRef = useRef(false)
+  const { remove, updateStatus, updateProgress } = useOngoingTransfers()
+  const { addCompletedTransfer } = useCompletedTransfers()
+  const { addNotification } = useNotification()
+  const chainflipSwaps = useMemo(() => getChainflipOngoingSwaps(ongoingTransfers), [ongoingTransfers])
+
+  const pollStatuses = useCallback(
+    async (swaps: StoredTransfer[]) => {
+      if (swaps.length === 0 || isPollingRef.current) return
+      isPollingRef.current = true
```
NunoAlexandre, line 120: what is this? `isPollingRef.current = true`

app/src/hooks/useFees.ts
```diff
@@ -317,6 +332,91 @@ export default function useFees(params: UseFeesParams) {
           break
         }
 
+        case 'ChainflipApi': {
+          if (!chainflipQuote || isChainflipQuoteError) {
+            setFees(null)
+            setIsBalanceSufficientForFees(true) // TODO: check if this is correct
```
NunoAlexandre, line 338: Have you checked the ToDo here? Do we need to create a follow up ticket for something regarding this?

app/src/hooks/useFees.ts
```diff
@@ -317,6 +332,91 @@ export default function useFees(params: UseFeesParams) {
...
+                  inDollars: feeTokenInDollars ? toHuman(fee.amount, token) * feeTokenInDollars : 0,
+                },
+              }
+            }),
+          )
+
+          const feeList: FeeDetails[] = [...chainflipfeeList]
+
+          if (sourceChain.network === 'Polkadot') {
+            const localTransferfeeToken = PolkadotTokens.DOT
```
NunoAlexandre, line 364:
```suggestion
            const localTransferFeeToken = PolkadotTokens.DOT
```

app/src/hooks/useFees.ts
```diff
@@ -447,3 +551,88 @@ const getCachedBridgingFee = async (): Promise<bigint> => {
...
+      abi: erc20Abi,
+      functionName: 'transfer',
+      args: [senderAddress, parseUnits(sourceTokenAmount.toString(), sourceToken.decimals)],
+    })
+
+    gas = await publicClient.estimateGas({
+      account: walletClient.account,
+      to: sourceToken.address as `0x${string}`, // Token smart contract address
+      data: data,
+    })
```
NunoAlexandre, line 586: Have we encountered an edge case where this is run? i.e, where the fee token is not ETH when sending from Ethereum?

app/src/hooks/useTransferForm.ts
```diff
@@ -387,6 +394,33 @@ const useTransferForm = () => {
...
+  // validate chainflip source token minimum amount to swap
+  useEffect(() => {
+    const validateChainflipSwap = async () => {
+      if (!sourceChain || !destinationChain || !sourceToken || !destToken || !sourceTokenAmount?.amount) {
+        setMinSwapAmountError('')
+        return
+      }
+
+      if (isChainflipSwap(sourceChain, destinationChain, sourceToken, destToken)) {
+        const srcChain = await getChainflipChain(sourceChain)
```
NunoAlexandre, line 406: for consistency, can we name this "source*" 🙏

app/src/hooks/useTransferProgress.ts
```diff
@@ -40,31 +48,42 @@ const getTransferProgress = (date: Date, direction: Direction, shouldStartProgre
   return Math.min(progress + 1, 90)
 }
 
+const getTransferDate = (transfer: StoredTransfer, isChainflipSwap: boolean) => {
+  // handles chainflip swaps
+  if (isChainflipSwap) return transfer.date
+  // Bridges, XCMs, Hydra swaps, etc
+  return transfer.finalizedAt ? transfer.finalizedAt : transfer.date
```
NunoAlexandre, line 55: it seems that we can simplify this to this?

```
return transfer.finalizedAt ?? transfer.date
```

app/src/utils/chainflip.ts
```diff
@@ -0,0 +1,357 @@
...
+import { getChainSpecificAddress } from './address'
+
+/** TYPES */
+export type AssetSymbol = 'DOT' | 'USDC' | 'USDT' | 'ETH' | 'FLIP' | 'BTC' | 'SOL'
+
+export type ChainflipChain = 'Ethereum' | 'Polkadot' | 'Assethub' | 'Arbitrum' | 'Bitcoin' | 'Solana'
+
+export type ChainflipFeeType = 'NETWORK' | 'INGRESS' | 'EGRESS' | 'BROKER' | 'BOOST' | 'REFUND'
+
+export type ChainflipQuote = RegularQuote | DCAQuote
```
NunoAlexandre, line 29:
```suggestion
export type AssetSymbol = 'DOT' | 'USDC' | 'USDT' | 'ETH' | 'FLIP' | 'BTC' | 'SOL'
export type ChainflipChain = 'Ethereum' | 'Polkadot' | 'Assethub' | 'Arbitrum' | 'Bitcoin' | 'Solana'
export type ChainflipFeeType = 'NETWORK' | 'INGRESS' | 'EGRESS' | 'BROKER' | 'BOOST' | 'REFUND'
export type ChainflipQuote = RegularQuote | DCAQuote
```

app/src/utils/chainflip.ts
```diff
@@ -0,0 +1,357 @@
...
+  response?: {
+    data?: {
+      message?: string
+    }
+  }
+}
+
+export enum ChainflipRefundStatus {
+  Refunded = 'refunded',
+  Partially = 'partially refunded',
```
NunoAlexandre, line 41: is there a way to know how much was refunded?

app/src/utils/chainflip.ts
```diff
@@ -0,0 +1,357 @@
...
+    if (route.from.chainId === sourceChain.chainId && route.pairs.some(([token, _]) => token.id === sourceToken.id)) {
+      chainsSet.add(route.to)
+    }
+  })
+
+  return Array.from(chainsSet)
+}
+
+/** Returns all Chainflip allowed destination tokens for a swap. */
+export const getChainflipSwapDestTokens = (
```
NunoAlexandre, line 80: Given that this function is already in a "chainflip" module, the redundancy of adding "Chainflip" to all of these names is rough. We could reference these functions using the module name for better namespacing, e.g.:

```
chainflip.getDestTokens(...)
```

app/src/utils/chainflip.ts
```diff
@@ -0,0 +1,357 @@
...
+
+  const tokensSet = new Set<Token>()
+  const route = chainflipRoutes.find(
+    route => route.from.chainId === sourceChain.chainId && route.to.chainId === destinationChain.chainId,
+  )
+  if (!route) return []
+
+  route.pairs.forEach(([srcT, destT]) => srcT.id === sourceToken.id && tokensSet.add(destT))
+
+  return Array.from(tokensSet)
```
NunoAlexandre, line 95: I would prefer a more functional structure to this. Plus expressions like "srcT.id === sourceToken.id && tokensSet.add(destT)` are borderline magic, it's not immediately understandable code.

```suggestion
 const desTokens = chainflipRoutes.find(route => 
   route.from.chainId === sourceChain.chainId && route.to.chainId === destinationChain.chainId,
  )?.pairs
  .filter(([srcT, _]) => srcT.id === sourceToken.id)
  .map(([_, destT]) => destT)


// return uniques
return [...new Set(destTokens)];
```

app/src/utils/chainflip.ts
```diff
@@ -0,0 +1,357 @@
...
+      amount,
+    })
+
+    //Find regular quote
+    const regularQuote = quotes.find(quote => quote.type === 'REGULAR')
+    //Find DCA quote
+    const dcaQuote = quotes.find(quote => quote.type === 'DCA')
+    if (!regularQuote && !dcaQuote) throw new Error('Chainflip quote not found.')
+
+    return regularQuote ? regularQuote : (dcaQuote ?? null)
```
NunoAlexandre, line 144:
```suggestion
    return regularQuote ?? dcaQuote ?? null
```

NunoAlexandre (commented):
This is huge 🙃 

Great work overall, Victor

app/src/components/Transfer.tsx
Victor-Salomon: @NunoAlexandre, I'm sorry my brain froze for a moment 😅. I tried your code and it works. I just got myself confused somehow 🥲

app/src/hooks/useChainflipTracker.ts
Victor-Salomon: This creates a reference flag, which we use to indicate a Promise is already running. It prevents starting a new one until the current one finishes, avoiding overlapping Promises.

app/src/hooks/useFees.ts
Victor-Salomon: Thanks for the reminder. 

I need to double-check with @jonidelv. It was introduced while I was working on the Chainflip implementation.

This should be correct.

app/src/hooks/useFees.ts
Victor-Salomon: ✅ It will always be DOT for local transfers on AH. 

Chainflip generates the deposit address on AH, and we do not support any direct swap from a parachain with Chainflip.

app/src/hooks/useFees.ts
jonidelv: Yes, in this case it’s basically a way of resetting/cleaning up the errors. We could create a function, maybe called reset or cleanUp, that does something like:
```
setFees(null)
setIsBalanceSufficientForFees(true)
```
But like we discussed, if there’s no scenario where fees + sending > user balance, then no worries, we don’t need it.

app/src/hooks/useFees.ts
Victor-Salomon: The else case is not an edge case. The gas fee token is always ETH, even when users swap USDT or USDC.

- For ETH swaps, we include the ETH amount directly in the `estimateGas` method and parse it with `parseEther()`
- For ERC-20 transfers (USDT/USDC), we call the token’s contract `transfer(...)` method and parse the ERC20 with`parseUnits()`. But the gas is still paid in ETH.

@Nuno, does it answer your question?

app/src/hooks/useTransferForm.ts
Victor-Salomon: hmm, `sourceChain` and `sourceAsset` names are already used here. 
`getChainflipChain()` converts a turtle registry chain to the chainflip format.
Same for `getChainflipAsset()`

I can rename them `chainflipSourceChain` and `chainflipAssetChain`.

app/src/hooks/useTransferProgress.ts
Victor-Salomon: hehe correct ✅

app/src/utils/chainflip.ts
NunoAlexandre: The list of routes and pairs we are handling is minimal so performance concerns is not a real issue in practice. In cases like that I prefer to prioritise code readability but that's also not a clear cut here either. I let you decide on this.

app/src/components/completed/TransactionDialog.tsx
```diff
@@ -14,6 +16,7 @@ interface TransactionDialogProps {
 }
 
 export default function TransactionDialog({ tx }: TransactionDialogProps) {
+  const isChainflipCheck = isChainflipSwap(tx.sourceChain, tx.destChain, tx.sourceToken, tx.destinationToken)
```
jonidelv, line 19: Since this is only used in the component declaration, could we move it inline there?`isChainflipCheck={isChainflipSwap(tx.sourceChain, tx.destChain, tx.sourceToken, tx.destinationToken)}`

app/src/components/completed/TransactionDialog.tsx
```diff
@@ -1,9 +1,11 @@
 import { colors } from '@velocitylabs-org/turtle-tailwind-config'
 import { cn, Icon, TokenLogo } from '@velocitylabs-org/turtle-ui'
 import { type CompletedTransfer, type TransferResult, TxStatus } from '@/models/transfer'
+import { isChainflipSwap } from '@/utils/chainflip'
 import { formatHours } from '@/utils/datetime'
 import { formatAmount, isSwap, toHuman } from '@/utils/transfer'
 import Account from '../Account'
+import ChainflipRefund from '../ChainflipRefund'
```
jonidelv, line 8: Could we use the @ alias here instead of a relative path? That way, if the file gets moved, the import won’t break: `@/components/ChainflipRefund`

app/src/hooks/useFees.ts
```diff
@@ -317,6 +332,91 @@ export default function useFees(params: UseFeesParams) {
           break
         }
 
+        case 'ChainflipApi': {
+          if (!chainflipQuote || isChainflipQuoteError) {
+            setFees(null)
+            setIsBalanceSufficientForFees(true)
+            break
```
jonidelv, line 339: I’d suggest using return here instead of break, since we’re already using return in other cases above.

app/src/utils/chainflip.ts
```diff
@@ -0,0 +1,354 @@
...
+  })
+
+  const assethub = chainflipChains.find(c => c.chain === 'Assethub')
+
+  // If no match is found, we default to Assethub as this is the only remaining chain supported
+  return chainsMatch ?? assethub
+}
+
+/** Returns a Turtle Chain matching with Chainflip chain. */
+export const toRegistryChain = (chainflipChain: ChainflipChain): Chain => {
```
jonidelv, line 247: To make the intent clearer, I’d rename this to `chainflipToRegistryChain`. That way, when it’s referenced in other places (like fees), it’s immediately clear that this is mapping from Chainflip to our registry.

app/src/models/transfer.ts
```diff
@@ -194,7 +195,15 @@ export type OnChainBaseEvents = {
...
 }
 
-type FeeDetailType = 'Execution fees' | 'Delivery fees' | 'Bridging fees' | 'Routing fees' | 'Swap fees'
+export type FeeDetailType =
+  | 'Execution fees'
+  | 'Delivery fees'
+  | 'Bridging fees'
+  | 'Routing fees'
+  | 'Swap fees'
+  | 'Broker fees'
```
jonidelv, line 204: What do you think about separating the values that are specific to Chainflip, and then making a union type with the ones we use in the app? Another option could be moving them to the bottom and adding a comment.

app/src/hooks/useChainflipTracker.ts
Victor-Salomon: No worries :) 

`useRefs` doesn’t trigger re-renders on update. It acts like a _lock_ that flags the ongoing promise and avoids overlapping async calls when we update a status. 

1. We first check `isPollingRef.current` _(l.119)_
2. If it’s already `true`, it means a previous poll is still running => we `return` early.
3. If it’s `false`, we flip it to `true`, fetch and update the status.  _(l.120)_
4. Set it back to `false` once everything is settled, in the `finally` block of the `try/finally`. _(l.144)_ _(no need to `catch` since I'm using a `Promise.allSettled` and handles errors as part of the `try`)_

app/src/utils/transfer.ts
```diff
@@ -127,17 +128,34 @@ const EXPLORERS = {
...
   subscan_kusama_ah: 'https://assethub-kusama.subscan.io/',
   subscan_kusama_bh: 'https://bridgehub-kusama.subscan.io/',
+
+  // Chainflip
+  chainflip: 'https://scan.chainflip.io/',
+}
+
+export const getChainflipExplorerLink = (swap: StoredTransfer, chainflipSwapId?: string): string | undefined => {
+  if (chainflipSwapId) return `${removeURLSlash(EXPLORERS.chainflip)}/swaps/${chainflipSwapId}`
+  if (swap.uniqueTrackingId) return `${removeURLSlash(EXPLORERS.chainflip)}/channels/${swap.uniqueTrackingId}`
```
jonidelv, line 138: Could we add this to analytics as well? There’s a helper in t`urtle/analytics/src/utils/explorer-links` to show the explorer link. We could check if it’s a swap made with Chainflip (the swap case is already handled by the isSwap prop)

app/src/utils/chainflip.ts
steoo: agree with @NunoAlexandre on this one, the `&& .add` is javascript being a very permissive mom

app/src/hooks/useFees.ts
```diff
@@ -317,6 +332,91 @@ export default function useFees(params: UseFeesParams) {
...
+          }
+          setLoading(true)
+
+          const chainflipfeeList: FeeDetails[] = await Promise.all(
+            chainflipQuote.includedFees.map(async fee => {
```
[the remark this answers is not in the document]
Victor-Salomon: Sure :) 
I'm adding a comment as it's not exported by Chainflip, and it's 3 extended Interfaces.

app/src/models/transfer.ts
Victor-Salomon: Makes sense :)

app/src/utils/transfer.ts
Victor-Salomon: sure

Victor-Salomon:
@NunoAlexandre it is ✅

PR #512, opened 2025-09-05
Chore: Bump Paraspell to v11 (migration)

jonidelv:
This PR upgrade Paraspell SDK from v10 to v11 and apply required code changes per migration guide.

### Description
https://paraspell.github.io/docs/migration/v10-to-v11.html
Removal of the multi Prefix, NODE ➡ CHAIN Renaming, new types definition and Decimal Abstraction Setting (abstractDecimals)

### Related Issue
https://linear.app/velocitylabs/issue/VEL-503/upgrade-paraspell-sdk-from-v10-v11


package.json
```diff
@@ -19,7 +19,7 @@
     "knip": "knip"
   },
   "devDependencies": {
-    "@biomejs/biome": "^2.1.2",
+    "@biomejs/biome": "2.1.3",
```
steoo, line 22: 👀

steoo (approved):
looks to me they're many renamings and not super big changes

package.json
jonidelv: Yes, I removed the `^` because was installing a newer version that introduced some changes that is better to update in a different PR, also we were using the 2.1.3 version in config https://github.com/velocitylabs-org/turtle/blob/9c60a77deba526a06ca1a63b094f83ac6ea0ae9f/biome.json#L2 that's why that's why the manor update

Victor-Salomon (approved):
Everything looks good. 
Nice move 🦾

NunoAlexandre (approved):

PR #520, opened 2025-09-15
Fix: AH max button for Chainflip Swaps

Victor-Salomon:
## Description
Max amount `btn` requires tweaking the `destination chain` & the `recipient` to match the local transfer expected params.

## Related Issue
Closes [VEL-508](https://linear.app/velocitylabs/issue/VEL-508/fix-assethub-max-amount-for-chainflip-swaps)

### DIFF
--- a/app/src/hooks/useTransferForm.ts
+++ b/app/src/hooks/useTransferForm.ts
@@ -265,12 +265,13 @@ const useTransferForm = () => {
           return
         }
 
+        const isSwap = isChainflipSwap(sourceChain, destinationChain, sourceToken, destToken)
         const recipient = getRecipientAddress(manualRecipient, destinationWallet) ?? destinationWallet.sender.address
         const params = {
           sourceChain,
-          destinationChain,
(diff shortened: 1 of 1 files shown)


noahjoeris:
Why do we need this? Can you explain?

Victor-Salomon:
> Why do we need this? Can you explain?

@noahjoeris Sure :) 

Chainflip generates a Vault Deposit Address, and needs us to handle manually the (local) transfer to this deposit address.
 
Because this is not directly supported by Chainflip, we sometimes need to tweak the form parameters during the swap process until the funds are deposited to the Vault: Example, the form destchain is ETH, but when we handle the deposit, the destChain needs to be AH since the Deposit Address is a Polkadot Vault. 

When _fetching the fees_ or clicking to the _maxButton_, we don't yet have access to the deposit address as it's generated later, and to leverage the already existing Polakdot/AH helpers from Paraspell we need to mix a bit this magic.

Let me know if this is unclear, we can jump on a call :)

noahjoeris (approved):

PR #521, opened 2025-09-16
feat: add hydration to moonbeam stablecoin flows

noahjoeris:
## Description

Adds USDC and USDT transfers from Hydration to Moonbeam. The other way around requires a fix by PAPI. So we need to update to latest version of PAPI and Paraspell SDK before enabling it.

Closes VEL-487


jonidelv (approved):
I'm creating a PR with the latest PAPI and Paraspell updates

PR #525, opened 2025-09-18
fix: Make ENS work again

noahjoeris:
## Description
ENS didn't work so I fixed it by switching from ethers to wagmi.

## Related Issue
Solves VEL-374


app/src/utils/address.ts
```diff
@@ -61,7 +61,7 @@ export const isValidAddressType = (address: string, types: AddressType[]): boole
  * @param address - The address string to be validated.
  * @returns True if the address is a valid Ethereum address, false otherwise.
  */
-const isValidEthereumAddress = (address: string): boolean => {
+export const isValidEthereumAddress = (address: string): boolean => {
   return isAddress(address)
 }
```
steoo, line 66: is it worth it to have a function that just returns another function? if it's matter of naming you can just `import as`

noahjoeris: I think it's a good abstraction. In case we switch library. Happened before. So lets keep it

app/src/components/Account.tsx
```diff
@@ -22,8 +22,10 @@ export default function Account({
   allowCopy = true,
   size = 14,
 }: AccountProps) {
-  const accountName = useLookupName(network, address)
-  const accountDisplay = accountName ? accountName : truncateAddress(address, 4, 4)
+  const { data: ensName } = useEnsName({
```
Victor-Salomon, line 25: Is this still working as expected with the `undefined` case? 

`const { data: ensName } = useEnsName({address: undefined})`

Is `useEnsName`still returning { data }?

app/src/components/ChainTrigger.tsx
```diff
@@ -39,13 +38,16 @@ export default function ChainTrigger({
   walletAddress,
 }: ChainTriggerProps) {
   // wallet and ens
-  const addressLookup = useLookupName(value?.network, walletAddress?.toLowerCase())
+  const { data: ensName } = useEnsName({
+    address: isValidEthereumAddress(walletAddress || '') ? (walletAddress as `0x${string}`) : undefined,
```
Victor-Salomon, line 42: Viem.js has a type for `0x${string}`

`import { type Address } from 'viem'`

Victor-Salomon (approved):
Nice job @noahjoeris 

Good to know, removing the following, resolves some Sentry issues such as `Failed to fetch`: 

`new ethers.CloudflareProvider()`

🔥

app/src/utils/papi.ts
```diff
@@ -44,10 +44,12 @@ export const extractPapiEvent = (event: TxEvent): OnChainBaseEvents | undefined
       // Get messageId from xcmPallet pallet (ex: DOT from Relay Chain to AH)
       if (section === 'XcmPallet' && method === 'Sent' && 'message_id' in data) {
         messageId = data.message_id.asHex()
+        console.log('messageId', messageId)
```
Victor-Salomon, line 47: @noahjoeris we have those logs now 😅

PR #530, opened 2025-09-23
Chore: Bump Paraspell SDK and related dependencies to v11.8.4

jonidelv:
Bump Paraspell SDK to latest version (v11.8.4)

## Related Issue
https://linear.app/velocitylabs/issue/VEL-514/update-paraspell-to-v1182


app/src/hooks/useFees.ts
```diff
@@ -597,9 +597,15 @@ const checkEip1559BalanceSufficiency = async (
...
 
 function getToken(location: TLocation | undefined, symbol: string) {
-  const token = location ? getTokenByLocation(location) : getTokenFromSymbol(symbol)
+  let token: Token | undefined
+  if (location) {
+    token = getTokenByLocation(location)
+  }
+  if (!token) {
+    token = getTokenFromSymbol(symbol)
+  }
```
NunoAlexandre, line 606:
```suggestion
  const token = location ? getTokenByLocation(location) ?? getTokenFromSymbol(symbol) : undefined
```

could be

NunoAlexandre (approved):
Giving a test drive

app/src/hooks/useFees.ts
jonidelv: Yes, just tha with the ternary approach (both the previous and this suggestion), if `getTokenByLocation(location)` fails, it just goes into the catch block and returns immediately without attempting a fallback. That’s why I kept the explicit `if (!token)` check, it ensures the symbol lookup still happens when the location lookup doesn’t return a token.
An alternative to the suggestion could be
```ts
const token = location && getTokenByLocation(location) || getTokenFromSymbol(symbol)
``` 
but I feel that it could be harder to read, but im open to suggestion, what do you think?

NunoAlexandre: you are right, let's keep it as is for the sake of readability

NunoAlexandre (approved):
The last bump seems good 👌

PR #531, opened 2025-09-23
refactor: migrate components from app to widget

noahjoeris:
## Description
This PR migrates components from the app to the widget. It adds new files and updates existing ones. This is done in order to use the widget inside the app.


steoo:
needs a branch out https://github.com/velocitylabs-org/turtle/pull/524

PR #538, opened 2025-09-26
Chore: Bump Paraspell SDK and related dependencies to v11.8.5

jonidelv:
Bump Paraspell SDK and related dependencies to v11.8.5


Victor-Salomon (approved):
✅

PR #553, opened 2025-11-05
Chore: Bump Paraspell SDK and related dependencies

jonidelv:
Paraspell has introduced new methods in the `paraspell/sdk` that are useful for us. The `getReceivableAmount` method automatically calculates the received amount. Additionally, the router includes `getMinTransferableAmount` and `getTransferableAmount` methods to determine if the user has enough balance to cover the transfer.


app/src/hooks/useOutputAmount.ts
```diff
@@ -149,3 +148,36 @@ export function useOutputAmount({
     isLoading: isLoading || isFetching || isLoadingChainflipQuote || !!loadingFees,
   }
 }
+
+function calcReceivableAmountManually({
```
Victor-Salomon: Not a big deal, but I found this naming a bit heavy 😅
How about `getReceivableAmount`?

No issue if you would rather keep it like this

jonidelv: Sounds good!

